PR #3, opened 2016-09-06
G4CMP-12: Improve interface between MeshElectricField and TriLinearIn…

tkreynolds:
The GetPotential and GetField functions have been moved out of G4CMPTriLinearInterp and into G4CMPMeshElectricField.

### DIFF
--- a/library/include/G4CMPMeshElectricField.hh
+++ b/library/include/G4CMPMeshElectricField.hh
@@ -18,6 +18,7 @@
 
 #include "G4CMPTriLinearInterp.hh"
 #include "G4ElectricField.hh"
+#include "G4ThreeVector.hh"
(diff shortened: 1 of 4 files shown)


ragnese:
Rebase develop and put new changes on top.

kelseymh:
I've never been happy with passing around pointers in place of properly dimensioned arrays or vectors; there's no way to ensure proper size for either filling or referencing.  Could you take a look at the alternate branch (G4CMP-12) where the Gradient() function returns a G4ThreeVector instead of a G4double*.  Also, that alternate branch also has all of the Geant4-specific "[4]" position+time arguments reduced to "[3]" positions.

ragnese:
Tyler is already on it.

PR #4, opened 2016-09-06
G4CMP-9

tkreynolds:
G4CMPTriLinearInterp now caches the electric field values. It checks against the previous tetrahedra index, and if the current point is not in a new tetrahedra, the cached values are returned. Otherwise, new field values are calculated and stored. Additionally, TriLinearInterp now uses vectors of arrays instead of vectors of vectors.


ragnese:
1. Move oldIdx inside of GetField
2. Two spaces for indent
3. Move `using point = ` to header and use in all function definitions and declarations.

PR #5, opened 2016-09-07
G4CMP-26

ragnese:
Hey Mike,

Can you do me a favor and look at the new function in G4CMPProcessUtils::ValidateSecondaryPosition(). In particular, do I need to actually do any local-to-global transformations? I think the position given by the track is in global coordinates, but I'm not positive about the G4VSolid::SurfaceNormal() coordinates. Are they global or local to the solid?

### DIFF
--- a/library/src/G4CMPProcessUtils.cc
+++ b/library/src/G4CMPProcessUtils.cc
@@ -882,6 +882,7 @@ G4ThreeVector G4CMPProcessUtils::ValidateSecondaryPosition(const G4ThreeVector& 
   // in the next volume. Instead, let's scoot a bit away from the edge.
   if (safety <= kCarTolerance) {
     G4ThreeVector norm = currentTrack->GetVolume()->GetLogicalVolume()->GetSolid()->SurfaceNormal(pos);
+    RotateToGlobalDirection(norm);
     secPos = pos + (safety - kCarTolerance * (1.001)) * norm;
   }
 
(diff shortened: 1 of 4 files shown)


library/src/G4CMPProcessUtils.cc
```diff
@@ -868,3 +880,22 @@ G4CMPProcessUtils::CreateChargeCarrier(G4int charge, G4int valley,
...
+  // If the step is near a boundary, create the secondary in the initial volume
+  G4Navigator* nav = G4TransportationManager::GetTransportationManager()->GetNavigatorForTracking();
+  // Safety is the distance to a boundary
+  G4double safety = nav->ComputeSafety(pos);
+  // Tolerance is the error in deciding which volume a track is in
+  G4double kCarTolerance = G4GeometryTolerance::GetInstance()->GetSurfaceTolerance();
+  // If the distance to an edge is within error, we might accidentally get placed
+  // in the next volume. Instead, let's scoot a bit away from the edge.
+  if (safety <= kCarTolerance) {
+    G4ThreeVector norm = currentTrack->GetVolume()->GetLogicalVolume()->GetSolid()->SurfaceNormal(pos);
```
kelseymh: The "norm" computed here should be rotated into global coordinates, for comparison with "pos."  Just do
`RotateToGlobalDirection(norm);`
which modifies 'norm' in situ.

PR #7, opened 2017-02-06
Changes to g4cmp_env.sh

gnilheb:
I had used the cmake workflow to build and install the library and then when I tried to run the examples I was immediately confronted with an exception saying that it could not locate the CrystalMaps directory. The proposed change fixes the problem.


ragnese:
Doesn't the new script set both LD_LIBRARY_PATH and DYLD_LIBRARY_PATH if you have neither set? That would be a problem.

Also, if you build with CMake you don't need either of those variables. I wonder if we want to or need to install different env scripts depending on how you build.

PR #10, opened 2022-11-03
Modified G4CMPKaplanQP to handle quasiparticle energy-dependent absor…

perry5334:
…ption efficiency

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,9 +507,11 @@ for the metal film (defined using the function
 | absorberEffSlope | Efficiency dependence vs. energy  | 0.              |
 | temperature      | Temperature of film         | 0.05e-3*K          |
 
-The `subgapAbsorption` parameter is optional.  It only applies if there is a
+The last five  parameters are optional. They only apply if there is a
 sensor involved which is sensitive to heat energy, in which case phonons
 below 2.*bandgap energy, and above 2.*absorberGap energy, should be treated
-as directly absorbed with the specified probability.  In this case, we
-recommend that user applications also set `/g4cmp/minEPhonons` to
-2.*absorberGap, to avoid excessive CPU from tracking unmeasurable phonons.
+as directly absorbed with the specified 'subgapAbsorption'. In this case,
+we recommend that user applications also set `/g4cmp/minEPhonons` to 2.*absorberGap,
+to avoid excessive CPU from tracking unmeasurable phonons. Quasiparticles
+in a sensor are subject to a baseline absorption efficiency 'absorbedEff'
+and energy-dependent efficiency modification 'absorberEffSlope'.
--- a/README.md
+++ b/README.md
@@ -500,8 +500,12 @@ for the metal film (defined using the function
 | phononLifetime   | Phonon lifetime in film at 2*bandgap | 242.*ps   |
 | phononLifetimeSlope | Lifetime dependence vs. energy | 0.29         |
 | vSound           | Speed of sound in film      | 3.26*km/s          |
-| subgapAbsorption | Probability to absorb energy below 2*bandgap | 0. |
-| absorberGap      | Bandgap of energy absorber  | 15e-6*eV (tungsten) |
+| lowQPLimit       | Minimum bandgap multiple    | 3.                 |
+| subgapAbsorption | Probability to absorb energy below 2*bandgap | 0.03 (optional) |
+| absorberGap      | Bandgap of "subgap absorber"  | 15e-6*eV (tungsten) |
+| absorberEff      | Quasiparticle absorption efficiency  | 0.3          |
+| absorberEffSlope | Efficiency dependence vs. energy  | 0.              |
+| temperature      | Temperature of film         | 0.05e-3*K          |
 
 The `subgapAbsorption` parameter is optional.  It only applies if there is a
 sensor involved which is sensitive to heat energy, in which case phonons
--- a/library/include/G4CMPKaplanQP.hh
+++ b/library/include/G4CMPKaplanQP.hh
@@ -88,15 +88,18 @@ public:
   void SetLowQPLimit(G4double value)          { lowQPLimit = value; }
   void SetSubgapAbsorption(G4double value)    { subgapAbsorption = value; }
   void SetAbsorberGap(G4double value)         { absorberGap = value; }
+  void SetAbsorberEff(G4double value)         { absorberEff = value; }
+  void SetAbsorberEffSlope(G4double value)    { absorberEffSlope = value; }
   void SetPhononLifetime(G4double value)      { phononLifetime = value; }
   void SetPhononLifetimeSlope(G4double value) { phononLifetimeSlope = value; }
   void SetVSound(G4double value)              { vSound = value; }
 
 protected:
-  // Check that the five required parameters are set to meaningful values
+  // Check that the seven required parameters are set to meaningful values
   G4bool ParamsReady() const {
     return (filmThickness > 0. && gapEnergy >= 0. && vSound > 0. &&
-	    phononLifetime > 0. && phononLifetimeSlope >= 0.);
+	    phononLifetime > 0. && phononLifetimeSlope >= 0. && 
+	    absorberEff > 0. && absorberEffSlope >= 0.);
   }
 
   // Compute the probability of a phonon reentering the crystal without breaking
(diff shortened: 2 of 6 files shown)


library/src/G4CMPKaplanQP.cc
```diff
@@ -75,7 +75,8 @@ G4double G4CMP::KaplanPhononQP(G4double energy,
 G4CMPKaplanQP::G4CMPKaplanQP(G4MaterialPropertiesTable* prop, G4int vb)
   : verboseLevel(vb), filmProperties(0), filmThickness(0.), gapEnergy(0.),
     lowQPLimit(3.), subgapAbsorption(0.), absorberGap(0.),
-    phononLifetime(0.), phononLifetimeSlope(0.), vSound(0.), temperature(0.) {
+    phononLifetime(0.), phononLifetimeSlope(0.), vSound(0.), temperature(0.), 
+    absorberEff(0.), absorberEffSlope(0.) {
```
kelseymh, line 79: Make sure you declare these data members in the "private:" section of the class definition file (G4CMPKaplanQP.hh).  I think I would move them up a couple of lines, and declare (and initialize) them right after "absorberGap".

library/src/G4CMPKaplanQP.cc
```diff
@@ -112,6 +115,8 @@ void G4CMPKaplanQP::SetFilmProperties(G4MaterialPropertiesTable* prop) {
     phononLifetime =      prop->GetConstProperty("phononLifetime");
     phononLifetimeSlope = prop->GetConstProperty("phononLifetimeSlope");
     vSound =              prop->GetConstProperty("vSound");
+    absorberEff =         prop->GetConstProperty("absorberEff");
+    absorberEffSlope =     prop->GetConstProperty("absorberEffSlope");
```
kelseymh, line 119: Please be careful with alignment.  I'd like all the "prop->" to line up.

library/src/G4CMPKaplanQP.cc
```diff
@@ -433,6 +440,21 @@ G4CMPKaplanQP::CalcQPAbsorption(G4double qpE,
   return EDep;
 }
 
+
+// Handle quasiparticle energy-dependent absorption efficiency
+
+G4double
+G4CMPKaplanQP::CalcQPEfficiency(G4double qpE) const{
```
kelseymh, line 447: Space before the opening brace.  I also think you can fit this whole declaration on one line.

library/src/G4CMPKaplanQP.cc
```diff
@@ -433,6 +440,21 @@ G4CMPKaplanQP::CalcQPAbsorption(G4double qpE,
...
+    G4double eff = absorberEff + absorberEffSlope * qpE;
+
+  if (verboseLevel>2) {
+    G4cout << " efficiency " << eff << G4endl;
```
kelseymh, line 451: Change "efficiency" to "CalcQPEfficiency", so we know what's being reported when we turn on this debugging output.

kelseymh (changes requested):

library/src/G4CMPKaplanQP.cc
kelseymh: Also, of course this function needs to be declared in the .hh file.

library/include/G4CMPKaplanQP.hh
```diff
@@ -88,15 +88,18 @@ public:
...
   void SetVSound(G4double value)              { vSound = value; }
 
 protected:
-  // Check that the five required parameters are set to meaningful values
+  // Check that the seven required parameters are set to meaningful values
   G4bool ParamsReady() const {
     return (filmThickness > 0. && gapEnergy >= 0. && vSound > 0. &&
-	    phononLifetime > 0. && phononLifetimeSlope >= 0.);
+	    phononLifetime > 0. && phononLifetimeSlope >= 0. && 
+	    absorberEff > 0. && absorberEffSlope >= 0.);
```
kelseymh, line 102: Take this (and the comment above) back out.  We want these new parameters to be optional, not required.  That way, the current behaviour remains the default.

library/include/G4CMPKaplanQP.hh
```diff
@@ -156,6 +162,8 @@ private:
   G4double lowQPLimit;		// Minimum X*delta to keep as a quasiparticle
   G4double subgapAbsorption;	// Probability to absorb energy below bandgap
   G4double absorberGap;		// Bandgap of secondary absorber material
+  G4double absorberEff;     // Quasiparticle absorption efficiency
```
kelseymh, line 165: Align the comment with its neighbors.

library/include/G4CMPKaplanQP.hh
```diff
@@ -156,6 +162,8 @@ private:
...
+  G4double absorberEffSlope;     // Energy dependence of qp absorption efficiency
```
kelseymh, line 166: Align the comment with its neighbors.

library/src/G4CMPKaplanQP.cc
```diff
@@ -98,6 +99,8 @@ void G4CMPKaplanQP::SetFilmProperties(G4MaterialPropertiesTable* prop) {
   if (!(prop->ConstPropertyExists("gapEnergy") &&
         prop->ConstPropertyExists("phononLifetime") &&
         prop->ConstPropertyExists("phononLifetimeSlope") &&
+        prop->ConstPropertyExists("absorberEff") &&
```
kelseymh, line 102: Take these two lines back out.  We want these parameters to be optional, not required.

library/src/G4CMPKaplanQP.cc
```diff
@@ -112,6 +115,8 @@ void G4CMPKaplanQP::SetFilmProperties(G4MaterialPropertiesTable* prop) {
     phononLifetime =      prop->GetConstProperty("phononLifetime");
     phononLifetimeSlope = prop->GetConstProperty("phononLifetimeSlope");
     vSound =              prop->GetConstProperty("vSound");
+    absorberEff =         prop->GetConstProperty("absorberEff");
```
kelseymh, line 118: Implement these two assignment statements using the ternary operator (`: ?`) as is done with lowQPlimit.  Use a default of 1. for the efficiency, and 0. for the slope.

kelseymh (changes requested):
This is definitely converging fast.  In addition to the changes in the class, please edit the header comments at the top of G4CMPPhononElectrode (both .hh and .cc) to mention these two new material properties.  Also, add these to the table at the end of the README.md file.

README.md
```diff
@@ -500,8 +500,12 @@ for the metal film (defined using the function
...
-| subgapAbsorption | Probability to absorb energy below 2*bandgap | 0. |
-| absorberGap      | Bandgap of energy absorber  | 15e-6*eV (tungsten) |
+| lowQPLimit       | Minimum bandgap multiple    | 3.                 |
+| subgapAbsorption | Probability to absorb energy below 2*bandgap | 0.03 (optional) |
+| absorberGap      | Bandgap of "subgap absorber"  | 15e-6*eV (tungsten) |
+| absorberEff      | Quasiparticle absorption efficiency  | 0.3          |
+| absorberEffSlope | Efficiency dependence vs. energy  | 0.              |
+| temperature      | Temperature of film         | 0.05e-3*K          |
 
 The `subgapAbsorption` parameter is optional.  It only applies if there is a
```
kelseymh, line 510: Would you mind expanding this?  "The last <N> parameters are options.  They only apply..."

kelseymh (changes requested):
Just one request in the README.md.  This is converging really well!

library/src/G4CMPKaplanQP.cc
```diff
@@ -433,6 +444,20 @@ G4CMPKaplanQP::CalcQPAbsorption(G4double qpE,
   return EDep;
 }
 
+
+// Handle quasiparticle energy-dependent absorption efficiency
+
+G4double G4CMPKaplanQP::CalcQPEfficiency(G4double qpE) const {
+    G4double eff = absorberEff + absorberEffSlope * qpE;
```
kelseymh, line 451: We'd like the slope to be a dimensionless quantity.  As written here, it has to have units of 1/energy.  Instead, I'd suggest we scale it by Delta (gapEnergy):

`G4double eff = absorberEff + absorberEffSlope * qpE/gapEnergy;`

kelseymh (changes requested):

library/src/G4CMPKaplanQP.cc
```diff
@@ -433,6 +444,20 @@ G4CMPKaplanQP::CalcQPAbsorption(G4double qpE,
...
+
+  if (verboseLevel>2) {
+    G4cout << " CalcQPEfficiency " << eff << G4endl;
+  }
+
+  return std::min(eff, 1.);
```
kelseymh, line 457: For the case of a negative slope, we ought to keep the efficiency non-negative as well:

`return std::max(0., std::min(eff, 1.));`

kelseymh (changes requested):

kelseymh:
I think this is ready for testing.  We can use the CDMS simulation, and set the new efficiency parameter with

`/CDMS/Material/AddPropertyValue G4_Al absorberEff 0.3`

kelseymh (approved):

PR #15, opened 2023-09-13
G4 cmp 362 impact sarkis yields

dfsadek:
For debugging and optimizing the implementation of IMPACT@TUNL ionization yield measurement result and the latest Sarkis model for ionization yield in Silicon.


kelseymh:
General comment: Please update ChangeHistory at the top with today's date, JIRA ticket name and a one-line description.

library/include/G4YieldDataReader.hh
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef G4YIELDDATAREADER_HH
+#define G4YIELDDATAREADER_HH
+
+#include "G4String.hh"
+#include "G4PhysicsLogVector.hh"
+#include "globals.hh"
+#include <vector>
+
+class G4YieldDataReader {
```
kelseymh, line 20: Please rename this to G4CMPYieldDataReader.  Where is the .cc file that implements this function?

kelseymh:
Please make sure that you add all of your new source files to library/CMakeLists.txt.

You can test this by building using CMake.

dfsadek:
> General comment: Please update ChangeHistory at the top with today's date, JIRA ticket name and a one-line description.

Done. Lives locally on my machine till we decide to commit the changes.

dfsadek:
> Please do not commit from Jupyter areas. Please remove the entire .ipynbcheckpoints directory.

I removed the .ipynbcheckpoints and .DS_store from G4CMP-362. I also added them in the .gitignore file just in case any of these files appear one more time. I doubt that could be a .ipynbcheckpoints on the master branch but I didn't want to touch the master. Could you please, double check?

library/include/G4YieldDataReader.hh
dfsadek: This was an attempt to create a data reader file for the ionization yield data, but I used a different method. This file is now deleted. Not committed yet.

dfsadek:
> Please make sure that you add all of your new source files to library/CMakeLists.txt.
> 
> You can test this by building using CMake.

I added my header and source files to library/CMakeLists.txt.

kelseymh:
> I removed the .ipynbcheckpoints and .DS_store from G4CMP-362. I also added them in the .gitignore file just in case any of these files appear one more time. I doubt that could be a .ipynbcheckpoints on the master branch but I didn't want to touch the master. Could you please, double check?

I don't see any changes to .gitignore.  Did you commit and push it?

I also do not see any change to ChangeHistory; please commit and push it.  If we need to change it again before merging, we can.  The repository must be complete and _ready to merge_.

ChangeHistory
```diff
@@ -6,6 +6,9 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn".
 
+2023-10-11  G4CMP-362 : Code review for the implementation of IMPACT@TUNL
```
kelseymh, line 9: The ChangeHistory message should not be talking about a "code review."  It should describe _what_ the tag includes.  Here "Implementation of IMPACT and Serkis yield models."

library/include/G4CMPImpactTunlNIEL.hh
```diff
@@ -0,0 +1,45 @@
...
+  // return the fraction of the specified energy which will be deposited as NIEL
+  // if an incoming particle with z1, a1 is stopped in the specified material
+  // a1 is in atomic mass units, energy in native G4 energy units.
+  //
+  virtual G4double 
+  PartitionNIEL(G4double energy, const G4Material *material, G4double Zin=0.,
+		G4double Ain=0.) const override;
+
+private:
+    const G4double B = 0.261; // this is the central value for B=0.261
```
kelseymh, line 39: The comment is redundant:  commenting "B=0.261" for an assignment statement `B=0.261` doesn't provide any information.  Explain here *what* the "B" parameter is, so the reader doesn't have to reverse engineer the code.

library/include/G4CMPSarkisNIEL.hh
```diff
@@ -0,0 +1,49 @@
...
+  //
+  virtual G4double 
+  PartitionNIEL(G4double energy, const G4Material * material, G4double Zin = 0.,
+		G4double Ain = 0.);
+    
+private:
+    const G4double SiZ = 14.0;
+    const G4double SiA = 28.09;
+    G4String fPath;                     // full path to the data file
+    std::ifstream inputFile;
```
kelseymh, line 43: Does this have to be a data member?  Or can the ifstream be instantiated in the function in the .cc file where the reading happens?  It is better to have fewer data members, and especially avoid big complex data members, unless they're necessary.

library/src/G4CMPImpactTunlNIEL.cc
```diff
@@ -0,0 +1,72 @@
...
+    return 1.;		// Won't get here after FATAL error
+  }
+  // Get effective Z,A of the material
+  const G4double Z = GetEffectiveZ(material);
+  const G4double A = GetEffectiveA(material) / (g / mole);
+    
+  // Check if the material is silicon or similar (within +-1 of Z and A of silicon)
+    
+  if (std::abs(Z - SiZ) <= 1.0 && std::abs(A - SiA) <= 1.0) {
+      G4Exception("G4CMPImpactTunlNIEL", "G4CMP1005", JustWarning,
```
kelseymh, line 57: It may be worth thinking about using an "if (first)" kind of condition, so that these warnings are only printed once per job, rather than being printed on every single call (every hit of every event).

kelseymh (commented):
Added comments about code and other file changes.

library/src/G4CMPImpactTunlNIEL.cc
```diff
@@ -0,0 +1,76 @@
...
+		  "No material passed to partition function");
+    return 1.;		// Won't get here after FATAL error
+  }
+  // Get effective Z,A of the material
+  const G4double Z = GetEffectiveZ(material);
+  const G4double A = GetEffectiveA(material) / (g / mole);
+    
+  // Check if the material is silicon or similar (within +-1 of Z and A of silicon)
+    
+  if (std::abs(Z - SiZ) <= 1.0 && std::abs(A - SiA) <= 1.0) {
```
dfsadek, line 51: This if statement should be revisited since it checks if the A and Z of the input material are within +/-1 of A and Z for Silicon. For SuperCDMS detectors it is either Silicon or not, so the +/-1 here is an arbitrary choice to allow for different possibilities for other detectors that use G4CMP. Should that be left as +/-1 or there is a better value to choose?

kelseymh: I would test only Z, I think.   If somebody used enriched silicon for a detector (like the Avogadro sphere), the model should still work.

library/src/G4CMPSarkisNIEL.cc
```diff
@@ -56,25 +50,31 @@ PartitionNIEL(G4double energy, const G4Material *material,G4double /*Zin = 0.*/,
...
     if (energy <= 3 * MeV) {
-        // Sarkis model function
-        
-        inputFile.open(fPath);                       // open the data file
-        lVector.Retrieve(inputFile, false);          // load the data from the text file
-        inputFile.close();
-        return lVector.Value(energy, idx);           // do the interpolation and return the yield value
+      // Sarkis model function
+      std::ifstream inputFile;
+      inputFile.open(fPath);                       // open the data file
```
kelseymh, line 62: You can do this in one line: `std::ifstream inputFile(fPath);`.

library/src/G4CMPSarkisNIEL.cc
```diff
@@ -56,25 +50,31 @@ PartitionNIEL(G4double energy, const G4Material *material,G4double /*Zin = 0.*/,
...
+      lVector.Retrieve(inputFile, false);          // load the data from the text file
+      inputFile.close();                           // close the data file
```
kelseymh, line 64: You can leave this out if you like (or not).  Since the input stream is now scoped, it will be deleted (and the file closed) automatically at `return`.

PR #16, opened 2023-10-23
G4 cmp 381

emichaud4:
Implement a method to find rotation matrix (both valleys and anti-valleys) from miller indices instead of Euler angles. Theses changes solves some of the problems from JIRA ticket : 
-G4CMP-382 (store anti-valleys)
-G4CMP-381 (rounding errors in Euler angles methods)
-G4CMP-375 (how IV scattering must use anti-valleys instead of multiplying by -1 50% of the time) 
-G4CMP-374 (Si rotation matrices convention)
-G4CMP-338 (Ge rotation matrices convention)

Need to looks at G4CMP-375 more specifically since the convention used may affect momentum after IV scattering.

### DIFF
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -6,6 +6,9 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn".
 
+2023-08-31  G4CMP-381 : Compute rotation matrices from valley's direction
+			instead of Euler angles. Add anti-valleys.
+
 2023-08-31  g4cmp-V08-04-01 Revert two recent changes to restore NTL energy.
 2023-08-31  G4CMP-380 : Withdraw G4CMP-373, causes zero-length/NaN steps.
 2023-08-24  G4CMP-377 : Withdraw G4CMP-359/360, retstore energy conservation.
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -6,8 +6,7 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn".
 
-2023-08-31  G4CMP-381 : Compute rotation matrices from valley's direction
-			instead of Euler angles. Add anti-valleys.
+2023-08-31  G4CMP-381 : Add valleys and anti-valleys from Miller indices.
 
 2023-08-31  g4cmp-V08-04-01 Revert two recent changes to restore NTL energy.
 2023-08-31  G4CMP-380 : Withdraw G4CMP-373, causes zero-length/NaN steps.
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -6,6 +6,7 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn".
 
+2023-08-31  g4cmp-V08-04-01 Revert two recent changes to restore NTL energy.
 2023-08-31  G4CMP-380 : Withdraw G4CMP-373, causes zero-length/NaN steps.
 2023-08-24  G4CMP-377 : Withdraw G4CMP-359/360, retstore energy conservation.
 2023-08-08  G4CMP-373 : Fix long-standing bug in NTL emission rate.
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -6,7 +6,8 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn".
 
-2023-08-31  G4CMP-381 : Add valleys and anti-valleys from Miller indices.
+2023-11-22  G4CMP-375 : Remove 50% momentum flip in G4CMPInterValleyScattering.
+2023-11-10  G4CMP-381 : Add valleys and anti-valleys from Miller indices.
 
 2023-08-31  g4cmp-V08-04-01 Revert two recent changes to restore NTL energy.
 2023-08-31  G4CMP-380 : Withdraw G4CMP-373, causes zero-length/NaN steps.
--- a/library/include/G4LatticeLogical.hh
+++ b/library/include/G4LatticeLogical.hh
@@ -38,7 +38,7 @@
 //		precompute valley inverse transforms
 // 20200608  Fix -Wshadow warnings from tempvec
 // 20210919  M. Kelsey -- Allow SetVerboseLevel() from const instances.
-// 20231017  E. Michaud -- Add 'AddValley(const G4ThreeVector&)' to compute rotation matrix from valley's direction
+// 20231017  E. Michaud -- Add 'AddValley(const G4ThreeVector&)' 
 
 #ifndef G4LatticeLogical_h
 #define G4LatticeLogical_h
@@ -208,7 +208,7 @@ public:
   // Transform for drifting-electron valleys in momentum space
   void AddValley(const G4RotationMatrix& valley);
   void AddValley(G4double phi, G4double theta, G4double psi);
-  void AddValley(const G4ThreeVector&);
+  void AddValley(const G4ThreeVector&, G4bool antival=false);
   void ClearValleys() {
     fValley.clear(); fValleyInv.clear();fValleyAxis.clear();
   }
--- a/library/include/G4LatticeLogical.hh
+++ b/library/include/G4LatticeLogical.hh
@@ -38,6 +38,7 @@
 //		precompute valley inverse transforms
 // 20200608  Fix -Wshadow warnings from tempvec
 // 20210919  M. Kelsey -- Allow SetVerboseLevel() from const instances.
+// 20231017  E. Michaud -- Add 'AddValley(const G4ThreeVector&)' to compute rotation matrix from valley's direction
 
 #ifndef G4LatticeLogical_h
 #define G4LatticeLogical_h
@@ -207,6 +208,7 @@ public:
   // Transform for drifting-electron valleys in momentum space
   void AddValley(const G4RotationMatrix& valley);
   void AddValley(G4double phi, G4double theta, G4double psi);
+  void AddValley(const G4ThreeVector&);
   void ClearValleys() {
     fValley.clear(); fValleyInv.clear();fValleyAxis.clear();
   }
--- a/library/src/G4LatticeLogical.cc
+++ b/library/src/G4LatticeLogical.cc
@@ -703,32 +703,33 @@ void G4LatticeLogical::AddValley(const G4ThreeVector& valleyDirVec) {
    double c=0;
    double d;
    double e;
+   double f;
     
    if (vx==0 && vy==0){
-       e=0;
+       f=0;
        b=1;
-       a=1;
+       a=0;
    }
    
    else {
-       e=sqrt(1-vz*vz);
-       b=vx/e;
-       a=-vy/e;
+       f=sqrt(1-vz*vz);
+       b=vx/f;
+       a=-vy/f;
    }
      
    d=-vz*b;
    e=vz*a;
    
-G4cout << "test12 : " << a << " " << b << " " << c << " " << d << " " << e << G4endl;
+G4cout << "test12 : " << a << " " << b << " " << c << " " << d << " " << e << " " << G4endl;
     
-//     G4ThreeVector colx(1,0,0);
-//     G4ThreeVector coly(0,1,0);
-//     G4ThreeVector colz(0,0,1);
+    G4ThreeVector colx(vx,vy,vz);
+    G4ThreeVector coly(a,b,c);
+    G4ThreeVector colz(d,e,f);
 //     //HepRep3x3 test1(aa,aa,aa,aa,aa,aa,aa,aa,aa);
-//     G4RotationMatrix test1(colx,coly,colz);
+    G4RotationMatrix test1(colx,coly,colz);
     
     
-//   G4cout << "test : " << test1 <<  G4endl;
+  G4cout << "test : " << test1 <<  G4endl;
     
 }
 
--- a/library/src/G4LatticeLogical.cc
+++ b/library/src/G4LatticeLogical.cc
@@ -694,16 +694,22 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
 // Store drifting-electron valley using valley's direction
 
 void G4LatticeLogical::AddValley(const G4ThreeVector& valleyDirVec) {
+    
+   
       
-   double vx=valleyDirVec.x();
-   double vy=valleyDirVec.y();
-   double vz=valleyDirVec.z();
-   double a;
-   double b;
-   double c=0;
-   double d;
-   double e;
-   double f;
+   long double vx=valleyDirVec.x();
+   long double vy=valleyDirVec.y();
+   long double vz=valleyDirVec.z();
+   long double normval=sqrt(vx*vx+vy*vy+vz*vz);
+   vx=vx/normval;
+   vy=vy/normval;
+   vz=vz/normval;
+   long double a;
+   long double b;
+   long double c=0;
+   long double d;
+   long double e;
+   long double f;
     
    if (vx==0 && vy==0){
        f=0;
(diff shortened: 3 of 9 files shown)


library/include/G4LatticeLogical.hh
```diff
@@ -38,6 +38,7 @@
 //		precompute valley inverse transforms
 // 20200608  Fix -Wshadow warnings from tempvec
 // 20210919  M. Kelsey -- Allow SetVerboseLevel() from const instances.
+// 20231017  E. Michaud -- Add 'AddValley(const G4ThreeVector&)' to compute rotation matrix from valley's direction
```
kelseymh, line 41: Please break comments at less than 80 columns.  See the lines above for examples.

library/src/G4LatticeLogical.cc
```diff
@@ -44,6 +44,7 @@
 // 20190906  M. Kelsey -- Default IV rate model to G4CMPConfigManager value.
 // 20200520  For MT thread safety, wrap G4ThreeVector buffer in function to
 //		return thread-local instance.
+// 20231017  E. Michaud -- Add 'AddValley(const G4ThreeVector&)' to compute rotation matrix from valley's direction
```
kelseymh, line 47: Please break comments at less than 80 columns. See the lines above for examples.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,148 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
   fValleyAxis.push_back(fValleyInv.back()*G4ThreeVector(1.,0.,0.));
 }
 
+// Store drifting-electron valley using valley's direction
+
+void G4LatticeLogical::AddValley(const G4ThreeVector& valleyDirVec) {
+      
+  // Find the rotation matrix elements. 
+  // The rows correspond to the rotated axis direction (first row is vx,vy,vz ; second row is a,b,c and third row is d,e,f)
```
kelseymh, line 699: Please break comments at less than 80 columns.   This includes all comments in this section and anywhere else.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,148 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
   fValleyAxis.push_back(fValleyInv.back()*G4ThreeVector(1.,0.,0.));
 }
 
+// Store drifting-electron valley using valley's direction
+
+void G4LatticeLogical::AddValley(const G4ThreeVector& valleyDirVec) {
+      
+  // Find the rotation matrix elements.
```
kelseymh, line 698: This description should be possible to make much shorter.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,148 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+  // a = -vxvy/e/vx
+  // a = -vy/e
+    
+    
+  // Finally, we compute d and e which is the same for every case :
+  // d=-vzb
+  // e=vza
+    
+    
+  double vx=valleyDirVec.x();
```
kelseymh, line 756: All of this code needs to be consolidated to make better use of G4ThreeVector and G4RotationMatrix functionality.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,148 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+  vx=vx/normval;
+  vy=vy/normval;
+  vz=vz/normval;
+  double a;
+  double b;
+  double c=0; // rotated y axis must stay in the X-Y plane
+  double d;
+  double e;
+  double f;
+  double antia; // find the corresponding anti-valley rotation matrix elements
```
kelseymh, line 769: You should not need to do the anti-valley stuff in parallel.  You should be able to call this function with the anti-valley direction and let it do the same work.

library/src/G4LatticeReader.cc
```diff
@@ -32,6 +32,7 @@
 // 20180815  F. Insulla -- Added IVRateQuad
 // 20181001  M. Kelsey -- Clarify IV rate parameters systematically
 // 20190704  M. Kelsey -- Add 'ivModel' to set default IV function by material
+// 20231017  E. Michaud -- Add 'valleyDir' to set rotation matrix with valley's direction instead of euler angles
```
kelseymh, line 35: Please break comments at less than 80 columns.

library/src/G4LatticeReader.cc
```diff
@@ -146,6 +147,7 @@ G4bool G4LatticeReader::ProcessToken() {
       fToken == "cij")      return ProcessStiffness();  // Elasticity element
   if (fToken == "emass")    return ProcessMassTensor();	// e- mass eigenvalues
   if (fToken == "valley")   return ProcessEulerAngles(fToken); // e- drift dirs
+  if (fToken == "valleydir")   return ProcessValleyDirection(); // miller indices valley dirs
```
kelseymh, line 150: Can you line up with the other lines, please?

library/src/G4LatticeReader.cc
```diff
@@ -392,12 +394,27 @@ G4bool G4LatticeReader::ProcessEulerAngles(const G4String& name) {
     G4cerr << "G4LatticeReader: Unknown rotation matrix " << name << G4endl;
     return false;
   }
-
```
kelseymh, line 395: Why did you delete this line?

library/src/G4LatticeReader.cc
```diff
@@ -392,12 +394,27 @@ G4bool G4LatticeReader::ProcessEulerAngles(const G4String& name) {
...
+
+G4bool G4LatticeReader::ProcessValleyDirection() {
+  G4double milleri=0., millerj=0., millerk=0.;
+  *psLatfile >> milleri >> millerj >> millerk;
+  if (verboseLevel>1)
+    G4cout << " ProcessValleyDirection "  << milleri << " " 
+	   << millerj << " " << millerk << G4endl;
+
+  G4ThreeVector valleyDirVec(milleri,millerj,millerk);
+  pLattice->AddValley(valleyDirVec);
```
kelseymh, line 413: This is where you should call `pLattice->AddValley(-valleyDirVec);` to get the anti-valley.

kelseymh (changes requested):
Let me know in Slack when you've had a chance to process my comments.  If you have questions, ask in Slack or replying to these comments.  Thank you!

library/src/G4LatticeReader.cc
```diff
@@ -392,12 +395,29 @@ G4bool G4LatticeReader::ProcessEulerAngles(const G4String& name) {
     G4cerr << "G4LatticeReader: Unknown rotation matrix " << name << G4endl;
     return false;
   }
-
+    
   G4double degOrRad = ProcessUnits("Angle");
   pLattice->AddValley(phi*degOrRad, theta*degOrRad, psi*degOrRad);
```
kelseymh, line 400: We need to be doing the antivalley here as well.  If the user's config.txt is using the old "valley" input, then we still need to compute the antivalley for them.  You shouldn't do it by modifying that AddValley function internally; instead, you should apply the appropriate change to the Euler angles (psi -> -psi?  Or if psi==0, something else?) and call AddValley a second time with the modified values.

kelseymh (changes requested):

library/src/G4LatticeLogical.cc
kelseymh: I do not see any changes to the description.  The description should not simply repeat the actual calculations in the code.  It should be explaining *why* the code is written, or *how* the code does the job.  Pretend you're at a whiteboard explaining it to another student.  What words would you say to them while writing down the equations?

library/src/G4LatticeLogical.cc
kelseymh: This still needs to be done.  If you're not sure how to do it, please say so here in these comments so we can work together to resolve.  Using the vector manipulations built into G4ThreeVector and G4RotationMatrix should make the code much more self-documenting.

kelseymh:
Please update ChangeHistory at the top with a one-line description, to make this branch ready for merging.

library/src/G4LatticeLogical.cc
emichaud4: I made a few changes such as shortening the explications, explaining that the matrix must be right-handed, the z axis must point upward for valleys and downward for anti-valleys, etc.). I'm not quite sure what else to change, I was trying to do something similar to ChargeCarrierTimeStep in branch G4CMP-374!

library/src/G4LatticeLogical.cc
emichaud4: I made some changes by using a tempvect(), .unit() and G4Rep3x3, I'm not sure how to consolidate that more. I like using a double for vx=tempvec().x() because we are using vx a few times in the lines below. Maybe I should change that?

library/src/G4LatticeReader.cc
emichaud4: Perfect, I'll add the anti-valleys there!

library/src/G4LatticeLogical.cc
kelseymh: Let's find time to go through together (maybe using your slides rather than the code directly) exactly what's going on in each stage.  My suspicion is that you "should" be able to build the rotation matrix similar to how Euler angles work, by computing a set of "rotate about X", "rotate about Z" actions based on the direction vector, rather than doing the algebra "by hand" for the nine matrix components.

However, I may be falling into the "round-off trap."  That is, if we do what I'm thinking, maybe we end up with a less precise matrix than if we consolidate the underlying arithmetic the way you are doing.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+   
+  else if (vx==0){ // special case if vx=0 but not vy
+      a=-1;
+      b=0;
+  }
+    
+  else if (vy==0){ // special case if vy=0 but not vx
+      a=0;
+      b=1;    
+  }
```
kelseymh, line 773: if (vy == 0) { a=0; b=1; }
else if (vx == 0) { a=-1; b=0; }
else {a = -vy/f; b = vx/f; }

a = (vy==0 ? 0 : vx==0 ? -1 : -vy/f);
b = (vy==0 ? 1 : vx==0 ?  0 : vx/f);

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+  
+  double a;
+  double b;
+  double c=0; // rotated y axis must stay in the X-Y plane
+  double d;
+  double e;
+  double f;
+
+  // rotated z axis points in the +z direction for valleys  
+  f=sqrt(1-vz*vz);
```
kelseymh, line 754: f = sqrt(1. - vz*vz) * (antival?-1:1);

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+  double vx=tempvec().x();
+  double vy=tempvec().y();
+  double vz=tempvec().z();
+  
+  double a;
+  double b;
+  double c=0; // rotated y axis must stay in the X-Y plane
+  double d;
+  double e;
+  double f;
```
kelseymh, line 751: Don't need to declare these in advance.  Just declare the variable in the one-line calculation.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+  }
+     
+  d=-vz*b;
+  e=vz*a;
+
+  G4Rep3x3 M;
+  M = G4Rep3x3( vx,vy, vz, 
+	a, b, c, 
+	d, e, f );
+
```
kelseymh, line 787: Delete

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+	a, b, c, 
+	d, e, f );
+      
+  // Store the valley's rotation matrix, its inverse and the valley's direction.
+  fValley.resize(fValley.size()+1);
+  fValley.back().set(M);
+  fValleyInv.push_back(fValley.back());
+  fValleyInv.back().invert();
+  fValleyAxis.push_back(tempvec());
+
```
kelseymh, line 794: Delete.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+  d=-vz*b;
+  e=vz*a;
+
+  G4Rep3x3 M;
+  M = G4Rep3x3( vx,vy, vz, 
+	a, b, c, 
+	d, e, f );
+      
+  // Store the valley's rotation matrix, its inverse and the valley's direction.
+  fValley.resize(fValley.size()+1);
```
kelseymh, line 789: fValley.back().setRows(vdir, G4ThreeVector(a,b,c), G4ThreeVector(d,e,f));

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+    
+  // Then we compute a (if vx=!0 and vx!=0) : 
+  // a = -bvy/vx
+  // a = -vxvy/e/vx
+  // a = -vy/e
+
+  // Finally, we compute d and e :
+  // d=-vzb
+  // e=vza 
+
```
kelseymh, line 737: Delete all of this derivation.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+  // First row is vx,vy,vz ; second row is a,b,c and third row is d,e,f
+    
+  // We chose the following convention :
+  //  - The rotated y axis must stay in the X-Y plane (c=0)
+  //  - Valley's rotated Z axis must have positive z component (f>0) 
+  //  - Anti-valley's rotated Z axis must have positive z component (f>0) 
+  //  - The three rotated axis should stay normalized :
+  //	vx^2 + vy^2 + vz^2 = 1 ; a^2 + b^2 = 1 ; d^2 + e^2 + f^2 = 1
+  //  - And be orthogonal :
+  //	ad + be = 0 ; avx + bvy = 0 ; dvx + evy + fvz = 0
```
kelseymh, line 709: Don't need the math, just a one line statement.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
...
+  //  - It must be a right hand coordinate system :
+  //	(vx,vy,vz) X (a,b,c) = (d,e,f) : d=-vzb ; e=vza ; f=vxb - vya
```
kelseymh, line 711: This is fine, but can skip showing d, e, f, expressions.  That comes from the cross-product itself.

library/src/G4LatticeLogical.cc
```diff
@@ -690,6 +691,109 @@ void G4LatticeLogical::AddValley(G4double phi, G4double theta, G4double psi) {
   fValleyAxis.push_back(fValleyInv.back()*G4ThreeVector(1.,0.,0.));
 }
 
+// Store drifting-electron valley using valley's direction
+
+void G4LatticeLogical::AddValley(const G4ThreeVector& valleyDirVec, G4bool antival) {
+      
+  // Find the rotation matrix elements. 
+  // The rows correspond to the rotated axis direction 
+  // First row is vx,vy,vz ; second row is a,b,c and third row is d,e,f
```
kelseymh, line 700: Maybe do this by "drawing" the matrix (could even start on preceding line).

kelseymh (changes requested):

ChangeHistory
```diff
@@ -6,6 +6,9 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn".
 
+2023-08-31  G4CMP-381 : Compute rotation matrices from valley's direction
+			instead of Euler angles. Add anti-valleys.
```
kelseymh, line 10: Do you think you could shorten this to one line?  If not, it's fine as it is.

library/src/G4LatticeReader.cc
```diff
@@ -392,12 +395,31 @@ G4bool G4LatticeReader::ProcessEulerAngles(const G4String& name) {
...
+  // Anti-valleys
+  pLattice->AddValley(phi*degOrRad+pi, -1*(theta*degOrRad+pi), -psi*degOrRad);
```
kelseymh, line 402: Isn't the second one equivalent to just -theta*degOrRad+pi (since +pi and -pi are equivalent)?

kelseymh (changes requested):

library/src/G4LatticeReader.cc
emichaud4: you are right! Changed it to : -theta*degOrRad+pi

ChangeHistory
emichaud4: Changed the comment to : Add valleys and anti-valleys from Miller indices.

Tell me if it's okay!

PR #17, opened 2023-11-22
removed the 50% momentum flip

emichaud4:
Removed the 50% momentum flip in G4CMPInterValleyScattering since we are now using anti-valleys instead of only using valleys and multiplying momentum by -1 50% of the time to account for anti-valleys.


kelseymh:
This looks fine,  Emile.  I'll do the merge when I'm back.from.travel.

PR #18, opened 2024-01-30
Multiple-path support for G4LATTICEDATA variable

jwi3:
As currently stands G4LatticeReader will only read a single path off of `G4LATTICEDATA` when searching for crystal map files. To avoid contaminating the central `G4LATTICEDATA` directory with custom crystal map files, G4LatticeReader::OpenFile was modified to first split fDataDir into separate paths (in user-entered order) and search each for the desired lattice file, rather than search an unmodified fDataDir. For users with only a single path defined on G4LATTICEDATA, this change will not affect the search for crystal map files.

Additionally, the main `CMakeLists.txt` file was modified slightly to avoid a failure condition with `git describe`; this change is documented in the [commit comments](https://github.com/kelseymh/G4CMP/commit/f1fd7257bca669f4d93e9736a3e0f20a020eaa78).


library/src/G4LatticeReader.cc
```diff
@@ -27,13 +27,13 @@
...
-//		for multiple optical IV potentials
+//    for multiple optical IV potentials
 // 20170821  Add transverse sound speed, L->TT fraction
 // 20180815  F. Insulla -- Added IVRateQuad
 // 20181001  M. Kelsey -- Clarify IV rate parameters systematically
 // 20190704  M. Kelsey -- Add 'ivModel' to set default IV function by material
-// 20231017  E. Michaud -- Add 'valleyDir' to set rotation matrix with valley's 
-//		 direction instead of euler angles
+// 20231017  E. Michaud -- Add 'valleyDir' to set rotation matrix with valley's
+//     direction instead of euler angles
```
kelseymh, line 36: Can you please add a dated comment below this line?  It helps us keep track of why changes were made.  Also, it seems that your text editor collapsed tabs (ASCII 0x09) to four spaces.  If you could fix the lines above this so that the text goes back to being aligned, I'd appreciate it.

library/src/G4LatticeReader.cc
```diff
@@ -110,9 +110,23 @@ G4bool G4LatticeReader::OpenFile(const G4String& filename) {
 
   G4String filepath = filename;
   psLatfile = new std::ifstream(filepath);
-  if (!psLatfile->good()) {			// Local file not found
+  if (!psLatfile->good()) {         // Local file not found
     filepath = fDataDir + "/" + filename;
-    psLatfile->open(filepath);			// Try data directory
+    char str[1024];
```
kelseymh, line 115: In this section, can you use `G4String` rather than <stdlib.h>?  G4String is equivalent to std::string, so you can use the various search functions etc.

library/src/G4LatticeReader.cc
```diff
@@ -327,13 +341,13 @@ G4bool G4LatticeReader::ProcessCrystalGroup(const G4String& name) {
 
   if (verboseLevel>1) {
     G4cout << " a " << a*lunit/angstrom << " b " << b*lunit/angstrom
-	   << " c " << c*lunit/angstrom << " Ang "
-	   << " alpha " << alpha/deg << " beta " << beta/deg
-	   << " gamma " << gamma/deg << " deg" << G4endl;
+     << " c " << c*lunit/angstrom << " Ang "
```
kelseymh, line 344: Here's another section where your editor mangled the indention.

kelseymh (commented):
Please update README.md to explain that G4LATTICEDATA may be set to a searchable path (and give the syntax).

Please update ChangeHistory wth a one-line dated entry at the top for this change.  I suggest using "PR-18" for the tag name on the line.

library/src/G4LatticeReader.cc
jwi3: Yet another thing I can blame Fortran for: getting me in the habit of always translating tabs to spaces. I disabled in my editor for this project, so hopefully this doesn't crop up again. I'll start double-checking before submitting.

kelseymh: Thank you very much!  I use Emacs for my editor, so it has language-specific formatting to maintain alignment.  That's been a huge help to me as I've learned Python :-/

I see you also went through and fixed a bunch of my misspellings :-)

jwi3: Having spell check in my editor has been more of a curse than a blessing, I can't ignore the red squiggly lines. It might be time for me to take the dive on Emacs too.

library/src/G4LatticeReader.cc
```diff
@@ -110,9 +111,25 @@ G4bool G4LatticeReader::OpenFile(const G4String& filename) {
 
   G4String filepath = filename;
   psLatfile = new std::ifstream(filepath);
-  if (!psLatfile->good()) {			// Local file not found
-    filepath = fDataDir + "/" + filename;
-    psLatfile->open(filepath);			// Try data directory
+  if (!psLatfile->good()) { 		// Local file not found
+    G4String data_tok = fDataDir;
```
kelseymh, line 115: It looks to me like you might want to use `G4Tokenizer` for this loop.
```
  G4Tokenizer split(fDataDir);
  G4String data_tok;
  while (!(data_tok = split(":")).empty()) {
    // do your stuff on data_tok here
  }
```

kelseymh (changes requested):
Jake, I tried to use the Web interface to resolve the merge conflict down at the bottom of the PR.  However, it won't let me commit the change back to your fork (which makes sense).  Essentially all you need to do, in the Web interface, is remove the conflict markers (<<<, ===, and >>> lines).  There should be a blank line below G4CMP-390, before the V08-05-00 line.  Once you've done that, we should be able to complete the merge onto the dummy "PR-18" branch I created.

jwi3:
I *believe* I got the changes right, but let me know if I missed anything

kelseymh (approved):

PR #21, opened 2024-02-23
G4CMP 395

imanataee:
In the energy-momentum calculation, we use a Taylor expansion, and the valley direction is where the energy hits its minima
The further the direction of the momentum from the valley, the bigger the inaccuracy will be
Currently, after the charge changes its direction due to scattering or transportation, the valley is not being changed.
To address this issue, a dynamic valley assignment is needed to ensure the valley the charge is assigned to is the closest valley

To do this, I propose FillParticleChange does not accept a valley input anymore; it calculates the valley from momentum and passes it to the 3-argument version of itself
The only place we use the 3-argument version of FillParticleChange is in LukeScattering, where I made it so that the valley gets calculated from Momentum

### DIFF
--- a/library/src/G4CMPVDriftProcess.cc
+++ b/library/src/G4CMPVDriftProcess.cc
@@ -90,17 +90,18 @@ G4CMPVDriftProcess::PostStepGetPhysicalInteractionLength(
 // Fill ParticleChange energy and mass for electron charge carrier momentum
 
 void 
-G4CMPVDriftProcess::FillParticleChange(G4int ivalley, const G4ThreeVector& p) {
+G4CMPVDriftProcess::FillParticleChange(const G4ThreeVector& p) {
   // Compute kinetic energy from momentum for electrons or holes
   G4double energy = 0.;
+  G4int valley = FindNearestValley(GetLocalDirection(p));
   if (IsElectron()){
-    energy = theLattice->MapPtoEkin(ivalley, GetLocalDirection(p));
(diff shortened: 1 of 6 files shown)


library/src/G4CMPVDriftProcess.cc
```diff
@@ -90,17 +90,18 @@ G4CMPVDriftProcess::PostStepGetPhysicalInteractionLength(
...
 
 void 
-G4CMPVDriftProcess::FillParticleChange(G4int ivalley, const G4ThreeVector& p) {
+G4CMPVDriftProcess::FillParticleChange(const G4ThreeVector& p) {
   // Compute kinetic energy from momentum for electrons or holes
   G4double energy = 0.;
+  G4int valley = FindNearestValley(GetLocalDirection(p));
   if (IsElectron()){
-    energy = theLattice->MapPtoEkin(ivalley, GetLocalDirection(p));
+    energy = theLattice->MapPtoEkin(valley, GetLocalDirection(p));
```
kelseymh, line 98: With this change, we are going to need to re-run the samples and check energy conservation again.

kelseymh (commented):
This seems very simple.  It also appears to me that being able to turn it off might possibly be useful (for example, for the studies you did in Talk 3).  Do you want to think about whether we want to support that functionality?  I think it could be done where you make the `FindNearestValley()` call, by putting that into an if condition.

Or do we think that having a "fixed valley" is sufficiently wrong that it shouldn't be allowed?

PR #31, opened 2024-10-11
G4CMP-407

emichaud4:
G4CMP-407 is ready to be merged to develop. In this branch, l0 is computed from physical constants. We also found that l0_h for Si is an order of magnitude off. This branch will fix this problem. If l0 and specified in config.txt, the value in config.txt takes precedence over the computed one. l0 is computed in G4LatticeLogical function ProcessL0(G4bool IsElec) which is called in G4LatticeManager::MakeLattice after reading through the config.txt.

### DIFF
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -6,6 +6,44 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
+2024-09-25  G4CMP-422 : Allow user to override IV rate model with UI command.
+
+2024-09-25  g4cmp-V09-01-00 Support sharing mesh fields across threads.
+2024-09-25  G4CMP-430 : Update CONTRIBUTING for ChangeHistory with big tasks.
+2024-09-25  G4CMP-244 : Reduce memory footprint of mesh electric field objects.
+2024-08-22  G4CMP-423 : Avoid wrong volume assignment in G4CMPHitMerging.
+
+2024-08-06  g4cmp-V09-00-00 Theoretically correct electron kinematics.
+2024-07-21  G4CMP-276 : Redesign electron kinematics to follow theory properly.
+2024-07-21  G4CMP-418 : Use electron conductivity mass consistently.
+2024-07-21  G4CMP-417 :	Restore Luke scattering rate from main branch
+2024-07-12  G4CMP-415 : Protect adaptive MFP (G4CMP-414) in zero-field case.
+2024-07-05  G4CMP-414 : Address compiler warnings/errors in debugging code.
+2024-07-04  G4CMP-413 : Adaptive TimeStepper MFP Respecting Velocity
+2024-07-04  G4CMP-412 : Adapting IV Scattering with Improved Kinematics.
+2024-07-03  G4CMP-408 : Fix Luke Scattering to be done in the Spherical frame.
+			Herring-Vogt transformation now uses electron mass. 
+			For electrons CurrentVelocity and GetKineticEnergy 
+			consider the band structure to return correct values.
+2024-07-03  G4CMP-409 : Add Initialization at the start of IVScattering Process.
+2024-05-14  G4CMP-398 : Removing K_HV and K_Valley from the code. Bugfix
+			for Luke Scattering changes introduced in G4CMP-369.
+2024-02-07  G4CMP-369 : Adapting Luke Scattering with the new kinematics.
+			Modifications to mfp.
+2023-07-09  G4CMP-352 : Use relativistic band-structure kinematics in
+			EqEMField and LatticeLogical. Use quasi-momentum for k.
+2023-02-10  G4CMP-351 : Change Electron to Effective Mass in LukeScattering.
+2022-12-10  G4CMP-349 : Ephonon being calculated from Erecoil in LukeScattering.
+2022-03-08  G4CMP-286 : Use true momentum, split out transforms in EqEMField.
+2021-10-01  G4CMP-285 : G4CMPTimeStepper should set electron dynamical mass.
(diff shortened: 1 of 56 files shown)


kelseymh:
Was this originally branched off of *develop*?  Or did you branch it from *G4CMP-120*?  If I merged onto develop, I don't want to pull your partial G4CMP-120 work over at the same time.  I know how to deal with that, but I just need to know the state.

ChangeHistory
```diff
@@ -50,6 +51,8 @@ ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 2024-05-02  G4CMP-378 : Correct expression for phonon-QP scattering energy.
 2024-05-02  G4CMP-344 : Improve KaplanQP memory usage with mutable buffers.
 
+2024-05-07  G4CMP-404 : Improve charge carriers-phonons scattering.
```
kelseymh, line 54: This suggests that you branched off of G4CMP-404, but not G4CMP-120 (see my comment above).

README.md
```diff
@@ -481,6 +482,10 @@ the crystal system.
 | ivQuadField | val | Minimum field for quadratic IV expression | V/m  |
 | ivQuadPower | exp | Exponent: rate = Rate*(E^2-Field^2)^(exp/2) | none |
 
+The keywords l0_e and l0_h are optional. If they are not specified in
+config.txt, they will be computed from other physical constants. If they
```
kelseymh, line 486: Can you give the formula here?  I know that plain text, or even Markdown, isn't great for math, but it would good to be specific.

library/src/G4LatticeLogical.cc
```diff
@@ -57,6 +57,7 @@
 //		<p> (transport momentum).
 // 20231017  E. Michaud -- Add 'AddValley(const G4ThreeVector&)'
 // 20240426  S. Zatschler -- Add explicit fallthrough statements to switch cases
+// 20240510  Add ProcessL0(G4 bool).
```
kelseymh, line 60: Is this correct?  And is it really from May?  Maybe you can make the comment more narrative?  e.g., "Add function to compute L0 from other parameters".

library/src/G4LatticeManager.cc
```diff
@@ -124,6 +124,8 @@ G4LatticeLogical* G4LatticeManager::LoadLattice(const G4Material* Mat,
   G4LatticeLogical* newLat = latReader.MakeLattice(latDir+"/config.txt");
   if (newLat) {
     newLat->SetDensity(Mat->GetDensity());
+    if (newLat->GetElectronScatter()==0) {newLat->SetElectronScatter(newLat->ComputeL0(true));}
+    if (newLat->GetHoleScatter()==0) {newLat->SetHoleScatter(newLat->ComputeL0(false));}
```
kelseymh, line 128: Maybe use "<= 0." for the conditionals?

kelseymh (changes requested):

ChangeHistory
emichaud4: If I remember correctly, I branched 407 off of 404 very shortly (less than a week) after branching 404 off of develop. I checked and there are no changes from 120 in 407. I also merge develop into 407 a few days ago to fix some the bugs.

README.md
emichaud4: I added a formula there. Let me know if the format is okay or if we need to specify what the physical constants are. Or maybe add a reference to Cabrera et al. (2010) paper?

library/src/G4LatticeLogical.cc
emichaud4: I changed it to "add function to compute L0 from other parameters". The date is the original date the function was added. Do you want to update the date to today's date?

library/src/G4LatticeManager.cc
emichaud4: Done!!

ChangeHistory
```diff
@@ -6,6 +6,7 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
+2024-10-11  G4CMP-407 : Calculate l0 from physical constants.
```
kelseymh, line 9: When you remove the G4CMP-404 line (below), update this to today's date.

kelseymh (changes requested):
With the fix I've requested in ChangeHistory, I think this is ready to be merged.

ChangeHistory
emichaud4: Done!

ChangeHistory
emichaud4: Done!

README.md
```diff
@@ -481,6 +482,13 @@ the crystal system.
 | ivQuadField | val | Minimum field for quadratic IV expression | V/m  |
 | ivQuadPower | exp | Exponent: rate = Rate*(E^2-Field^2)^(exp/2) | none |
 
+The keywords l0_e and l0_h are optional. If they are not specified in
+config.txt, they will be computed from other physical constants: 
+
+l0 = pi*hbar^4*rho/(2*m^3*Dac^2)
```
kelseymh, line 488: Nice; thank you!  I think for purposes of the README file, it might be easier for users to spell out the parameter names.  Maybe something like
```
l0 = pi*hbar*4*density / (2*mass^3*acDeform^2)
```

library/src/G4LatticeLogical.cc
kelseymh: Thanks, Emile!  I didn't realize you started this that far back; sorry!

kelseymh (changes requested):

README.md
emichaud4: I'll update that part very soon! ComputeCanada is very very slow and I can't even compile the changes (I've been waiting for 1 hour and reconnecting to the server does not help).

emichaud4: Done!

kelseymh (approved):

PR #32, opened 2025-01-29
Merge G4CMP-447 to G4CMP-317

nttenpas:
This adds the FillParticleChange() method that updates the phonon's wavevector, momentum direction, and velocity double after a reflection and after QET emission.

### DIFF
--- a/library/src/G4CMPPhononBoundaryProcess.cc
+++ b/library/src/G4CMPPhononBoundaryProcess.cc
@@ -249,9 +249,9 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
 					: "on surface") << G4endl;
   }
 
-  trackInfo->SetWaveVector(reflectedKDir);
-  particleChange.ProposeVelocity(v);
-  particleChange.ProposeMomentumDirection(vdir);
+  // Update trackInfo wavevector and particleChange's group velocity and momentum direction
+  // reflectedKDir is in global coordinates here - no conversion needed
+  FillParticleChange(particleChange, aTrack, reflectedKDir)
 }
 
 
--- a/library/src/G4CMPPhononBoundaryProcess.cc
+++ b/library/src/G4CMPPhononBoundaryProcess.cc
@@ -29,6 +29,7 @@
 // 20220905  G4CMP-310 -- Add increments of kPerp to avoid bad reflections.
 // 20220910  G4CMP-299 -- Use fabs(k) in absorption test.
 // 20240718  G4CMP-317 -- Initial implementation of surface displacement.
+// 20250124  G4CMP-447 -- Use FillParticleChange() to update wavevector and Vg.
 
 #include "G4CMPPhononBoundaryProcess.hh"
 #include "G4CMPAnharmonicDecay.hh"
--- a/library/src/G4CMPPhononBoundaryProcess.cc
+++ b/library/src/G4CMPPhononBoundaryProcess.cc
@@ -216,7 +216,6 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
   }
 
   G4ThreeVector vdir = theLattice->MapKtoVDir(mode, reflectedKDir);
-  G4double v = theLattice->MapKtoV(mode, reflectedKDir);
 
   if (verboseLevel>2) {
     G4cout << "\n New wavevector direction " << reflectedKDir
@@ -251,7 +250,7 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
 
   // Update trackInfo wavevector and particleChange's group velocity and momentum direction
   // reflectedKDir is in global coordinates here - no conversion needed
-  FillParticleChange(particleChange, aTrack, reflectedKDir)
+  FillParticleChange(particleChange, aTrack, reflectedKDir);
 }
 
 
--- a/library/src/G4CMPPhononBoundaryProcess.cc
+++ b/library/src/G4CMPPhononBoundaryProcess.cc
@@ -216,7 +216,11 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
     refltype = "diffuse";
   }
 
-  G4ThreeVector vdir = theLattice->MapKtoVDir(mode, reflectedKDir);
+  // Update trackInfo wavevector and particleChange's group velocity and momentum direction
+  // reflectedKDir is in global coordinates here - no conversion needed
+  FillParticleChange(particleChange, aTrack, reflectedKDir);
+
+  G4ThreeVector vdir = *particleChange.GetMomentumDirection();
 
   if (verboseLevel>2) {
     G4cout << "\n New wavevector direction " << reflectedKDir
@@ -248,10 +252,6 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
 					: place==kOutside ? "OUTSIDE"
 					: "on surface") << G4endl;
   }
-
-  // Update trackInfo wavevector and particleChange's group velocity and momentum direction
-  // reflectedKDir is in global coordinates here - no conversion needed
-  FillParticleChange(particleChange, aTrack, reflectedKDir);
 }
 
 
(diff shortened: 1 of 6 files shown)


library/src/G4CMPProcessUtils.cc
```diff
@@ -192,6 +193,26 @@ void G4CMPProcessUtils::FindLattice(const G4VPhysicalVolume* volume) {
...
 
+// Fill ParticleChange wavevector and group velocity for a given wavevector
+// Wavevector is expected to be in the global coordinate frame
+void G4CMPProcessUtils::FillParticleChange(G4ParticleChange& particleChange,
+            const G4Track& track, const G4ThreeVector& wavevector) const {
+  // Get phonon mode from track
+  G4int mode = GetPolarization(track);
+
+  // Get Vg from global wavevector
+  G4ThreeVector vDir = theLattice->MapKtoVDir(mode, wavevector);
```
kelseymh, line 204: Need to pass in rotated (local coordinates) wavevector to `MapKtoVDir()` and `MapKtoV()`.  Since `wavevector` is const, you can do this with `GetLocalDirection(wavevector)` as the argument to the Map functions.

library/src/G4CMPProcessUtils.cc
```diff
@@ -192,6 +193,26 @@ void G4CMPProcessUtils::FindLattice(const G4VPhysicalVolume* volume) {
...
+  G4double v = theLattice->MapKtoV(mode, wavevector);
```
kelseymh, line 205: As above.

library/src/G4CMPProcessUtils.cc
```diff
@@ -192,6 +193,26 @@ void G4CMPProcessUtils::FindLattice(const G4VPhysicalVolume* volume) {
...
+
+  // Update trackInfo and particleChange
+  auto trackInfo = G4CMP::GetTrackInfo<G4CMPPhononTrackInfo>(track);
+  trackInfo->SetWaveVector(wavevector);
+  particleChange.ProposeVelocity(v);
+  vDir = RotateToGlobalDirection(vDir);
```
kelseymh, line 211: This shouldn't have compiled.  The `RotateToGlobal` modify the argument, and do not return a value.  You would just do `RotateToGlobalDirection(vDir);`

library/src/G4PhononScattering.cc
```diff
@@ -89,20 +90,11 @@ G4VParticleChange* G4PhononScattering::PostStepDoIt( const G4Track& aTrack,
...
-  RotateToGlobalDirection(vdir);
+  // Update particleChange with new wavevector and Vg
+  FillParticleChange(aParticleChange, aTrack, newK)
 
   if (verboseLevel>1)
-    G4cout << " new vgrp " << vgrp << " along " << vdir << G4endl;
-
-  aParticleChange.ProposeMomentumDirection(vdir);
-  aParticleChange.ProposeVelocity(vgrp);
+    G4cout << " new vgrp " << particleChange.GetVelocity() << " along " << *particleChange.GetMomentumDirection() << G4endl;
```
kelseymh, line 97: Use `aParticleChange` here.  The other object (`particleChange`) is a pointer, and a separate data member.  Don't ask...

kelseymh (changes requested):

library/src/G4CMPProcessUtils.cc
nttenpas: Ah! This might be a bug in the specular reflection code. In the specular reflection code, we pass `reflectedKDir` in the global direction (G4CMPPhononBoundaryProcess.cc:316), and it's unrotated in that function when doing the mapping (G4CMPUtils.cc:207 & G4LatticePhysical.cc:136). Maybe I am misunderstanding?

I'll change these to the local coordinates.

library/src/G4CMPProcessUtils.cc
nttenpas: You beat me here :) This has been fixed

library/src/G4PhononScattering.cc
nttenpas: This has been fixed as well!

library/src/G4CMPProcessUtils.cc
nttenpas: Just updated this to local coordinates.

library/src/G4CMPProcessUtils.cc
nttenpas: Updated this to local coordinates.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -215,8 +216,11 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
     refltype = "diffuse";
   }
 
-  G4ThreeVector vdir = theLattice->MapKtoVDir(mode, reflectedKDir);
-  G4double v = theLattice->MapKtoV(mode, reflectedKDir);
+  // Update trackInfo wavevector and particleChange's group velocity and momentum direction
+  // reflectedKDir is in global coordinates here - no conversion needed
+  FillParticleChange(particleChange, aTrack, reflectedKDir);
+
+  G4ThreeVector vdir = *particleChange.GetMomentumDirection();
```
kelseymh, line 223: This doesn't look right.  `particleChange` should be a non-const reference (hence the use of "."), not a pointer.

library/src/G4CMPPhononElectrode.cc
```diff
@@ -188,10 +189,13 @@ ProcessReflection(const G4Track& track, const G4Step& step,
...
 					  reflectedKDir, surfNorm));
 
-  if (verboseLevel>1)
-    G4cout << " Phonon reflected from QET toward " << reflectedKDir << G4endl;
+  FillParticleChange(particleChange, track, reflectedKDir);
 
-  particleChange.ProposeMomentumDirection(reflectedKDir);
+  if (verboseLevel>1){
+    G4cout << " Phonon reflected from QET with: Wavevector " << reflectedKDir
+    << ", Momentum " << *particleChange.GetMomentumDirection()
```
kelseymh, line 196: This should not have a pointer-dereference, and it shouldn't have compiled.  Just `particleChange.Get...`.

library/src/G4PhononScattering.cc
```diff
@@ -89,20 +90,11 @@ G4VParticleChange* G4PhononScattering::PostStepDoIt( const G4Track& aTrack,
...
-  RotateToGlobalDirection(vdir);
+  // Update particleChange with new wavevector and Vg
+  FillParticleChange(aParticleChange, aTrack, newK);
 
   if (verboseLevel>1)
-    G4cout << " new vgrp " << vgrp << " along " << vdir << G4endl;
-
-  aParticleChange.ProposeMomentumDirection(vdir);
-  aParticleChange.ProposeVelocity(vgrp);
+    G4cout << " new vgrp " << aParticleChange.GetVelocity() << " along " << *aParticleChange.GetMomentumDirection() << G4endl;
```
kelseymh, line 97: `aParticleChange` is an object, not a pointer.  Should not have "*", and it shouldn't have compiled this way.

kelseymh (changes requested):

library/src/G4CMPPhononBoundaryProcess.cc
nttenpas: I think that the return from .GetMomentumDirection() is a pointer. I referenced: https://apc.u-paris.fr/~franco/g4doxy/html/classG4ParticleChange.html#9af2bfbe537e72aad5b8e2b76aa2a49a

I originally did `particleChange.GetMomentumDirection();` and it was printing out the memory address of the momentum. I did run a test job, with this setup and it looks good. I can check again.

kelseymh: You're right; I just looked at G4ParticleChange.hh.  That must date back to *very* early in Geant4 development (like 1998 or so).  It _should_ be getting returned as a const reference.

PR #33, opened 2025-01-30
Merge G4CMP-453  to G4CMP-317

nttenpas:
This branch introduces utility functions for getting the current track and current touchable pointers from the G4EventManeger. These utilities are used to do the coordinate rotations in PhononVelocityIsInward.

### DIFF
--- a/library/src/G4CMPUtils.cc
+++ b/library/src/G4CMPUtils.cc
@@ -23,6 +23,7 @@
 #include "G4CMPDriftElectron.hh"
 #include "G4CMPDriftHole.hh"
 #include "G4CMPElectrodeHit.hh"
+#include "G4CMPGeometryUtils.hh"
 #include "G4CMPTrackUtils.hh"
 #include "G4LatticePhysical.hh"
 #include "G4ParticleDefinition.hh"
@@ -206,10 +207,10 @@ G4bool G4CMP::PhononVelocityIsInward(const G4LatticePhysical* lattice,
                                      const G4ThreeVector& waveVector,
                                      const G4ThreeVector& surfNorm) {
   // MapKtoVDir requires local direction for the wavevector
-  G4ThreeVector vDir = lattice->MapKtoVDir(mode, GetLocalDirection(waveVector));
+  G4ThreeVector vDir = lattice->MapKtoVDir(mode, G4CMP::GetLocalDirection(waveVector));
 
   // Compare group velocity and surface normal in global coordinates
-  RotateToGlobalDirection(vDir);
+  G4CMP::RotateToGlobalDirection(vDir);
   return vDir.dot(surfNorm) < 0.0;
 }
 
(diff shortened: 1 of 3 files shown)


library/src/G4CMPUtils.cc
```diff
@@ -199,12 +215,20 @@ G4ThreeVector G4CMP::LambertReflection(const G4ThreeVector& surfNorm) {
...
+// waveVector and surfNorm need to be in global coordinates
 
 G4bool G4CMP::PhononVelocityIsInward(const G4LatticePhysical* lattice,
                                      G4int mode,
                                      const G4ThreeVector& waveVector,
                                      const G4ThreeVector& surfNorm) {
-  G4ThreeVector vDir = lattice->MapKtoVDir(mode, waveVector);
+  // Get touchable for coordinate rotations
+  const G4VTouchable* touchable = GetCurrentTouchable();
+
```
kelseymh, line 226: Missed something last time.  The touchable pointer "could" be null, which should be checked.  It should never happen in normal track processing.

You could be primitive and just do something like
```
   if (!touchable) {
    G4cerr << "ERROR PhononVelocityIsInward() got null touchable." << G4endl;
    return false;
  }
```
Or you could use `G4Exception` and abort the event.

kelseymh (changes requested):

library/src/G4CMPUtils.cc
nttenpas: Should this be done in the GetCurrentTouchable function itself? That way we don't need to put the exit on every call, and it's handled in one spot. Maybe something like:

`const G4VTouchable* G4CMP::GetCurrentTouchable() {
  G4Track* track = GetCurrentTrack();
  if (!track) <do track exception>

  const G4VTouchable* touchable = track->GetTouchable();
  if (!touchable) <do exception for touchable>

  return touchable
}
`

Though I guess this assumes all calls want to exit the same way. I'm happy to keep the original function, and just put exception in PhononVelocityIsInward.

kelseymh: There are cases where the null touchable is valid (if the G4Track being processed happens to be a primary, and we're seeing it in a function called from StackingAction, for instance).  I'd rather have the lowest level functions just "do their thing", and let the high level function decide whether a null pointer is a Bad Thing or not.

library/src/G4CMPUtils.cc
```diff
@@ -199,12 +216,25 @@ G4ThreeVector G4CMP::LambertReflection(const G4ThreeVector& surfNorm) {
...
+  if (!touchable) {
+    G4Exception("G4CMP::PhononVelocityIsInward", "G4CMPUtils001",
+		FatalException, "Current track does not have valid touchable!");
+  }
```
kelseymh, line 231: Insert a ```return false;``` line inside the if block.  If we change the exception severity in the future, the code could print the error and then continue, so we don't want a subsequent segfault.

kelseymh (changes requested):

PR #34, opened 2025-02-05
Merge G4CMP-459 to G4CMP-317

nttenpas:
The G4CMP-459 branch adds functions for handling the edge cases in the displacement loop. This adds a function for reflecting the wavevector on a neighbor's surface, as well as a function for getting the step position on an edge case.

### DIFF
--- a/library/include/G4CMPPhononBoundaryProcess.hh
+++ b/library/include/G4CMPPhononBoundaryProcess.hh
@@ -58,11 +58,11 @@ protected:
   G4ThreeVector GetLambertianVector(const G4ThreeVector& surfNorm,
 				    G4int mode) const;
 
-  void AdjustSurfaceDirectionAtEdge(G4ThreeVector& kTan,
-            const G4ThreeVector& stepLocalPos) const;
-
   void AdjustPositionAtEdge(G4ThreeVector& stepLocalPos,
             G4ThreeVector& kTan, const G4double stepSize) const;
+
+  void AdjustSurfaceDirectionAtEdge(const G4ThreeVector& stepLocalPos,
+            G4ThreeVector& kTan) const;
 
 private:
   G4CMPAnharmonicDecay* anharmonicDecay;
--- a/library/include/G4CMPPhononBoundaryProcess.hh
+++ b/library/include/G4CMPPhononBoundaryProcess.hh
@@ -58,9 +58,12 @@ protected:
   G4ThreeVector GetLambertianVector(const G4ThreeVector& surfNorm,
 				    G4int mode) const;
 
+
+  // Modify stepLocalPos in place. Calls AdjustSurfaceDirectionAtEdge.
   void AdjustPositionAtEdge(G4ThreeVector& stepLocalPos,
             G4ThreeVector& kTan, const G4double stepSize) const;
 
+  // Modifies kTan in place.
   void AdjustSurfaceDirectionAtEdge(const G4ThreeVector& stepLocalPos,
             G4ThreeVector& kTan) const;
 
--- a/library/include/G4CMPPhononBoundaryProcess.hh
+++ b/library/include/G4CMPPhononBoundaryProcess.hh
@@ -13,8 +13,7 @@
 // 20181010  J. Singh -- Use new G4CMPAnharmonicDecay for boundary decays
 // 20181011  M. Kelsey -- Add LoadDataForTrack() to initialize decay utility.
 // 20220906  M. Kelsey -- Encapsulate specular reflection in function.
-// 20250204  G4CMP-459 -- Add GetReflectionOnEdge() for reflecting k on edge cases.
-// 20250205  G4CMP-459 -- Add GetEdgePosition() for surface displacement edge cases.
+// 20250204  N. Tenpas -- Support reflection displacement search at hard corners.
 
 #ifndef G4CMPPhononBoundaryProcess_h
 #define G4CMPPhononBoundaryProcess_h 1
@@ -59,10 +58,10 @@ protected:
   G4ThreeVector GetLambertianVector(const G4ThreeVector& surfNorm,
 				    G4int mode) const;
 
-  G4ThreeVector GetReflectionOnEdge(const G4ThreeVector& stepLocalPos,
-            const G4ThreeVector& kTan) const;
+  void AdjustSurfaceDirectionAtEdge(G4ThreeVector& kTan,
+            const G4ThreeVector& stepLocalPos) const;
 
-  G4ThreeVector GetEdgePosition(const G4ThreeVector& stepLocalPos,
+  void AdjustPositionAtEdge(G4ThreeVector& stepLocalPos,
             const G4ThreeVector& kTan, const G4double stepSize) const;
 
 private:
--- a/library/include/G4CMPPhononBoundaryProcess.hh
+++ b/library/include/G4CMPPhononBoundaryProcess.hh
@@ -13,6 +13,8 @@
 // 20181010  J. Singh -- Use new G4CMPAnharmonicDecay for boundary decays
 // 20181011  M. Kelsey -- Add LoadDataForTrack() to initialize decay utility.
 // 20220906  M. Kelsey -- Encapsulate specular reflection in function.
+// 20250204  G4CMP-459 -- Add GetReflectionOnEdge() for reflecting k on edge cases.
+// 20250205  G4CMP-459 -- Add GetEdgePosition() for surface displacement edge cases.
 
 #ifndef G4CMPPhononBoundaryProcess_h
 #define G4CMPPhononBoundaryProcess_h 1
@@ -56,6 +58,12 @@ protected:
 
   G4ThreeVector GetLambertianVector(const G4ThreeVector& surfNorm,
 				    G4int mode) const;
+
+  G4ThreeVector GetReflectionOnEdge(const G4ThreeVector& stepLocalPos,
+            const G4ThreeVector& waveVector) const;
+
+  G4ThreeVector GetEdgePosition(const G4ThreeVector& stepLocalPos, 
+            const G4ThreeVector& waveVector) const;
 
 private:
   G4CMPAnharmonicDecay* anharmonicDecay;
--- a/library/include/G4CMPPhononBoundaryProcess.hh
+++ b/library/include/G4CMPPhononBoundaryProcess.hh
@@ -60,10 +60,10 @@ protected:
 				    G4int mode) const;
 
   G4ThreeVector GetReflectionOnEdge(const G4ThreeVector& stepLocalPos,
-            const G4ThreeVector& waveVector) const;
+            const G4ThreeVector& kTan) const;
 
   G4ThreeVector GetEdgePosition(const G4ThreeVector& stepLocalPos,
-            const G4ThreeVector& waveVector, const G4double stepSize) const;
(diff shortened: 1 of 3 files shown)


library/include/G4CMPPhononBoundaryProcess.hh
```diff
@@ -57,6 +59,12 @@ protected:
   G4ThreeVector GetLambertianVector(const G4ThreeVector& surfNorm,
 				    G4int mode) const;
 
+  G4ThreeVector GetReflectionOnEdge(const G4ThreeVector& stepLocalPos,
+            const G4ThreeVector& kTan) const;
+
+  G4ThreeVector GetEdgePosition(const G4ThreeVector& stepLocalPos,
```
kelseymh, line 65: Instead of returning by value, you could name this function something like `AdjustPositionToEdge()`, pass in stepLocalPos as a non-const reference, and overwrite the value for output.

That reduces creating/deleting a temporary object every time it's called.

library/include/G4CMPPhononBoundaryProcess.hh
```diff
@@ -57,6 +59,12 @@ protected:
   G4ThreeVector GetLambertianVector(const G4ThreeVector& surfNorm,
 				    G4int mode) const;
 
+  G4ThreeVector GetReflectionOnEdge(const G4ThreeVector& stepLocalPos,
```
kelseymh, line 62: Instead of returning by value, you could name this function something like AdjustSurfaceDirectionAtEdge(), pass in kTan as a non-const reference, and overwrite the value for output.

That reduces creating/deleting a temporary object every time it's called.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -327,7 +328,15 @@ GetReflectedVector(const G4ThreeVector& waveVector,
 
     // FIXME: Find point on surface nearest to stepLocalPos, and reset
     surfAdjust = solid->DistanceToIn(stepLocalPos, -newNorm);	// Get distance along normal from new position back to detector surface
-    stepLocalPos -= surfAdjust * newNorm;			// Adjust position to be back on detector surface
+
+    // If surfAdjust > 1, we stepped off an edge and need to correct
+    if (surfAdjust > 1) {
+      stepLocalPos = GetEdgePosition(stepLocalPos, reflectedKDir, stepSize);
+      kTan = GetReflectionOnEdge(stepLocalPos, kTan);
+    } else {
```
kelseymh, line 336: See my comments in the .hh file for updating value within function.  Take a look at `RotateToGlobalDirection()` for an example of such an _in situ_ update.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -410,3 +419,50 @@ GetLambertianVector(const G4ThreeVector& surfNorm, G4int mode) const {
 
   return reflectedKDir;
 }
+
+
+// Get the position on the edge of two surfaces
+
+G4ThreeVector G4CMPPhononBoundaryProcess::
```
kelseymh, line 426: See my comment in the .hh file about passing in non-const reference and modifying in place.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -410,3 +419,50 @@ GetLambertianVector(const G4ThreeVector& surfNorm, G4int mode) const {
...
+  // Step remaining distance along surface
+  edgePos += (stepSize - surfAdjust) * newKTan;
+
+  return edgePos;
+}
+
+
+// Reflect "surface mode" phonon at the edge cases
+
+G4ThreeVector G4CMPPhononBoundaryProcess::
```
kelseymh, line 453: See my comment in the .hh file about passing in non-const reference and modifying in place.

library/include/G4CMPPhononBoundaryProcess.hh
```diff
@@ -57,6 +58,12 @@ protected:
   G4ThreeVector GetLambertianVector(const G4ThreeVector& surfNorm,
 				    G4int mode) const;
 
+  void AdjustSurfaceDirectionAtEdge(G4ThreeVector& kTan,
+            const G4ThreeVector& stepLocalPos) const;
+
+  void AdjustPositionAtEdge(G4ThreeVector& stepLocalPos,
+            const G4ThreeVector& kTan, const G4double stepSize) const;
+
```
kelseymh, line 66: If it was me, I would have kept the previous order of arguments, so that they're the same in both functions, just make the appropriate argument non-const.  

I understand you did it this way so that "modified" argument comes first, but I think switching the order is going to be more confusing.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -245,7 +246,7 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
 	   << "\n from " << surfacePoint << "\n   to " << stepPos << G4endl;
 
     G4ThreeVector stepLocal = GetLocalPosition(stepPos);
-    G4VSolid* solid = aStep.GetPreStepPoint()->GetPhysicalVolume()->GetLogicalVolume()->GetSolid();
+    solid = aStep.GetPreStepPoint()->GetPhysicalVolume()->GetLogicalVolume()->GetSolid();
```
kelseymh, line 249: I'm very confused here.  Why isn't this declared with a type?  That only works if "solid" is a class data member, which it is not (at least, it's not in the .hh file).

kelseymh (changes requested):

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -327,7 +328,15 @@ GetReflectedVector(const G4ThreeVector& waveVector,
 
     // FIXME: Find point on surface nearest to stepLocalPos, and reset
     surfAdjust = solid->DistanceToIn(stepLocalPos, -newNorm);	// Get distance along normal from new position back to detector surface
-    stepLocalPos -= surfAdjust * newNorm;			// Adjust position to be back on detector surface
+
+    // If surfAdjust > 1, we stepped off an edge and need to correct
+    if (surfAdjust > 1) {
```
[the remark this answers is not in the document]
nttenpas: I think testing vs 1mm is reasonable. My thinking was that our step size is < 1mm, so if surfAdjust is ever > 1, we certainly stepped off an edge.

library/include/G4CMPPhononBoundaryProcess.hh
```diff
@@ -13,6 +13,8 @@
 // 20181010  J. Singh -- Use new G4CMPAnharmonicDecay for boundary decays
 // 20181011  M. Kelsey -- Add LoadDataForTrack() to initialize decay utility.
 // 20220906  M. Kelsey -- Encapsulate specular reflection in function.
+// 20250204  G4CMP-459 -- Add GetReflectionOnEdge() for reflecting k on edge cases.
+// 20250205  G4CMP-459 -- Add GetEdgePosition() for surface displacement edge cases.
```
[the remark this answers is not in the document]
nttenpas: Updated to: Support reflection displacement search at hard corners.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -327,7 +328,15 @@ GetReflectedVector(const G4ThreeVector& waveVector,
 
     // FIXME: Find point on surface nearest to stepLocalPos, and reset
     surfAdjust = solid->DistanceToIn(stepLocalPos, -newNorm);	// Get distance along normal from new position back to detector surface
-    stepLocalPos -= surfAdjust * newNorm;			// Adjust position to be back on detector surface
+
+    // If surfAdjust > 1, we stepped off an edge and need to correct
```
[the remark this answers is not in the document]
nttenpas: This has been updated: If surfAdjust is large we have stepped off a hard corner and need correction

library/include/G4CMPPhononBoundaryProcess.hh
nttenpas: Updated! Thank you for this, that makes sense and I'll keep this in mind in future work.

library/include/G4CMPPhononBoundaryProcess.hh
nttenpas: Updated!

library/src/G4CMPPhononBoundaryProcess.cc
nttenpas: I have fixed my interface so I don't need to push and pull every change before checking compilation :-). Sorry about this

library/include/G4CMPPhononBoundaryProcess.hh
nttenpas: Agreed! This has been changed back

library/include/G4CMPPhononBoundaryProcess.hh
```diff
@@ -57,6 +58,12 @@ protected:
   G4ThreeVector GetLambertianVector(const G4ThreeVector& surfNorm,
 				    G4int mode) const;
 
+  void AdjustPositionAtEdge(G4ThreeVector& stepLocalPos,
+            G4ThreeVector& kTan, const G4double stepSize) const;
```
kelseymh, line 62: In this one, I think kTan should be a const reference.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -410,3 +418,47 @@ GetLambertianVector(const G4ThreeVector& surfNorm, G4int mode) const {
 
   return reflectedKDir;
 }
+
+
+// Modify the position for a full step off a hard corner
+
+void G4CMPPhononBoundaryProcess::
+AdjustPositionAtEdge(G4ThreeVector& stepLocalPos, 
+        G4ThreeVector& kTan, const G4double stepSize) const {
```
kelseymh, line 427: const reference, eh?

library/include/G4CMPPhononBoundaryProcess.hh
kelseymh: Oh!  I missed that bit; I thought the functions were called sequentially in the main loop.  In that case, you're quite right.  It would probably be good to add an explicit comment for each function in the .hh file, _saying_ that the function will modify "xyz" in place.

nttenpas: That's a good idea. I updated the comments in the header file as well as the source code on the latest commit.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -410,3 +418,47 @@ GetLambertianVector(const G4ThreeVector& surfNorm, G4int mode) const {
...
+  // Get normal at current position
+  G4VSolid* solid = GetCurrentVolume()->GetLogicalVolume()->GetSolid();
+  G4ThreeVector currNorm = solid->SurfaceNormal(stepLocalPos);
+
+  // Get bordering surface's normal
+  G4ThreeVector edgePos = stepLocalPos - 1.*mm * currNorm;
+  G4ThreeVector newNorm = solid->SurfaceNormal(edgePos);
+
+  // Reflect vector against new normal
+  G4double kPerp = kTan * newNorm;
```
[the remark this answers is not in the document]
nttenpas: That makes sense. That change is on the 8d3ea29 commit.

kelseymh:
This looks ready to go, Nolan.  Coordinate with Wade, if you can.

kelseymh (approved):

PR #37, opened 2025-02-12
G4 cmp 460

dfsadek:
Fix Lindhard-Robinson ionization yield model.


kelseymh (approved):
Thank you for fixing this, David.  My preference, for consistency with Geant4, would be that A would be passed in "properly" with the units that G4 expects, and then we would divide it out within the function.  But the current interface, where A is supposed to be passed "like an integer," has been in place long enough that user applications are using it.

PR #38, opened 2025-02-14
Provide references for ionization yield models

dfsadek:
Provide references for ionization yield models.


kelseymh (approved):
Sorry I overlooked this PR.  I'll get it merged onto develop now.

PR #41, opened 2025-03-25
G4CMP463 -> G4CMP317

nttenpas:
Added macro commands for surface displacement step size and max number of steps taken.

### DIFF
--- a/library/include/G4CMPConfigManager.hh
+++ b/library/include/G4CMPConfigManager.hh
@@ -46,7 +46,7 @@
 // 20220921  G4CMP-319:  Add temperature setting for use with QP sensors.
 // 20221117  G4CMP-343:  Add option flag to preserve all internal phonons.
 // 20240506  G4CMP-371:  Add flag to keep or discard below-minimum track energy.
-// 20250305  G4CMP-463:  Add parameter for phonon surface displacement step size.
+// 20250325  G4CMP-463:  Add parameter for phonon surface step size & limit.
 
 #include "globals.hh"
 #include <iosfwd>
@@ -96,6 +96,7 @@ public:
   static G4double GetHATrapIonMFP()      { return Instance()->hATrapIonMFP; }
   static G4double GetTemperature()       { return Instance()->temperature; }
   static G4double GetPhononSurfStepSize()  { return Instance()->pSurfStepSize; }
+  static G4double GetPhononSurfStepLimit()  { return Instance()->pSurfStepLimit; }
 
   static const G4String& GetLatticeDir() { return Instance()->LatticeDir; }
   static const G4String& GetIVRateModel() { return Instance()->IVRateModel; }
@@ -107,6 +108,7 @@ public:
   static void SetMaxChargeBounces(G4int value) { Instance()->ehBounces = value; }
   static void SetMaxPhononBounces(G4int value) { Instance()->pBounces = value; }
   static void SetPhononSurfStepSize(G4double value) { Instance()->pSurfStepSize = value; }
+  static void SetPhononSurfStepLimit(G4int value) { Instance()->pSurfStepLimit = value; }
   static void SetMaxLukePhonons(G4int value) { Instance()->maxLukePhonons = value; }
   static void SetSurfaceClearance(G4double value) { Instance()->clearance = value; }
   static void SetMinStepScale(G4double value) { Instance()->stepScale = value; }
@@ -167,6 +169,7 @@ private:
   G4int ehBounces;	// Maximum e/h reflections ($G4CMP_EH_BOUNCES)
   G4int pBounces;	// Maximum phonon reflections ($G4CMP_PHON_BOUNCES)
   G4int maxLukePhonons; // Approx. Luke phonon limit ($G4MP_MAX_LUKE)
+  G4int pSurfStepLimit;  // Phonon surface displacement step limit ($G4CMP_PHON_SURFLIMIT).
   G4String version;	// Version name string extracted from .g4cmp-version
   G4String LatticeDir;	// Lattice data directory ($G4LATTICEDATA)
   G4String IVRateModel;	// Model for IV rate ($G4CMP_IV_RATE_MODEL)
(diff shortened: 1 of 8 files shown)


library/include/G4CMPConfigManager.hh
```diff
@@ -94,6 +95,8 @@ public:
   static G4double GetHDTrapIonMFP()      { return Instance()->hDTrapIonMFP; }
   static G4double GetHATrapIonMFP()      { return Instance()->hATrapIonMFP; }
   static G4double GetTemperature()       { return Instance()->temperature; }
+  static G4double GetPhononSurfStepSize()  { return Instance()->pSurfStepSize; }
+  static G4double GetPhononSurfStepLimit()  { return Instance()->pSurfStepLimit; }
```
kelseymh, line 99: I think this should be returning `G4int`.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -64,7 +66,12 @@
 
 G4CMPPhononBoundaryProcess::G4CMPPhononBoundaryProcess(const G4String& aName)
   : G4VPhononProcess(aName, fPhononReflection), G4CMPBoundaryUtils(this),
-    anharmonicDecay(new G4CMPAnharmonicDecay(this)), stepSize(50*um) {;}
+    anharmonicDecay(new G4CMPAnharmonicDecay(this)), stepSize(0*um), nStepLimit(0) {
+
```
kelseymh, line 70: Remove the extra blank line.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -295,14 +302,21 @@ GetReflectedVector(const G4ThreeVector& waveVector,
...
   G4ThreeVector oldkPerpV = kPerpV;
   G4ThreeVector oldstepLocalPos = stepLocalPos;
 
+  // Set default stepSize based on solid bounding limits
+  if (stepSize == 0) {
+    G4ThreeVector pmin(0,0,0);
+    G4ThreeVector pmax(0,0,0);
+    solid->BoundingLimits(pmin, pmax);
+    stepSize = (pmax - pmin).mag() / nStepLimit;
+  }
```
kelseymh, line 318: Great idea, but this may have to change to be general.  If a user's geometry has multiple devices, it would need to be recomputed each time.  Maybe instead of fetching the ConfigManager value in the ctor, you fetch it here, and if _it's_ zero, do the bounding calculation?

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -311,13 +325,15 @@ GetReflectedVector(const G4ThreeVector& waveVector,
       << ", newNorm = " << newNorm
       << ", reflectedKDir = " << reflectedKDir
       << ", kPerpV (kPerpMag * newNorm) = " << kPerpV
-      << ", kTan (reflectedKDir - kPerpV) = " << kTan << G4endl;
+      << ", kTan (reflectedKDir - kPerpV) = " << kTan
+      << ", surfaceStepSize = " << stepSize
```
kelseymh, line 329: You could use `G4BestUnit(stepSize)` here (from `#include "G4UnitsTable.hh"`), or you could explicitly choose mm or um or something.

kelseymh (changes requested):

library/src/G4CMPPhononBoundaryProcess.cc
nttenpas: I am making the change to:

`  // Initialize stepSize and max step limit from config manager
  G4CMPConfigManager* config = G4CMPConfigManager::Instance();
  stepSize = config->GetPhononSurfStepSize();
  nStepLimit = config->GetPhononSurfStepLimit();
  // Set default stepSize based on solid bounding limits
  if (stepSize == 0) {
    G4ThreeVector pmin(0,0,0);
    G4ThreeVector pmax(0,0,0);
    solid->BoundingLimits(pmin, pmax);
    stepSize = (pmax - pmin).mag() / nStepLimit;
  }`
  
But I admit I am not sure I understand why this changes anything. Won't the stepSize var get filled in the ctor, and be saved as a member variable throughout the job? I guess I am not clear when the ctor gets invoked.

nttenpas: But I admit I am not sure I understand why this changes anything. Won't the stepSize var get filled in the ctor, and be saved as a member variable throughout the job? I guess I am not clear when the ctor gets invoked.

kelseymh: Yes, that's exactly right, Nolan!  Consider the following sequence, assuming that the user has a complex geometry, consisting of one little chip device _and_ one big honkin' detector.

1) User doesn't set anything via macro, so G4CMPConfigManager has stepSize=0.
2) BoundaryProcess fetches the config value in it's ctor, so in the ctor, stepSize=0.
3) First track goes through the little chip, makes phonons.
4) BoundaryProcess::PostStepDoIt() gets called, does its thing, ends up in the surface displacement code.
5) stepSize=0, so code gets the bounding box of the little chip, resets stepSize=100*nm.
6) Second track goes through the big detector, makes phonons.
7) BoundaryProcess ends up in surface displacement code again.
8) stepSize=100*nm (from (5) above), so that gets used with the big honkin' detector.  Oops!

nttenpas: Oh! I see exactly what you're saying! I misunderstood the comment at first, and thought my original configuration was causing multiple computations. 

I am still trying to fully understand C++ memory management, but it sounds like I can put the following in the ctor:

```G4CMPConfigManager* config = G4CMPConfigManager::Instance();
nStepLimit = config->GetPhononSurfStepLimit();```

and this right before the displacement loop:

```G4CMPConfigManager* config = G4CMPConfigManager::Instance();
stepSize = config->GetPhononSurfStepSize();```
along with the default if-statement.

This way, the max step limit is always saved, and we don't need to get it every time. Each time we enter the loop, the config manager can use the deallocated memory space from the initial ctor call. 

Does this sound right, or am I overcomplicating things?

Comment (kelseymh) [reply library/src/G4CMPPhononBoundaryProcess.cc:318]: Your code implementation is fine.  If you want to, you can still keep the stepSize call back to ConfigManager in the BoundaryProcess ctor _also_ (just for symmetry with the nStepLimit call).

We can talk about the memory managment issue separately :-)

File: library/include/G4CMPPhononBoundaryProcess.hh [hunk 5]
```diff
@@ -53,7 +54,7 @@ protected:
 
   G4ThreeVector GetReflectedVector(const G4ThreeVector& waveVector, 
 				                           G4ThreeVector& surfNorm, G4int mode,
-                                   G4ThreeVector& surfacePoint) const;
+                                   G4ThreeVector& surfacePoint);
```
kelseymh, line 57: Does this compile and run without the 'const'?  If it gets called _from_ a const function, it'll fail.  You can avoid that by making the stepSize data member "mutable" (so const functions are allowed to change it).  If this is working as is, you don't need to change anything.

nttenpas: Yes, this is compiling and running. The function that calls GetReflectedVector,  DoReflection, is also non-const. 

Looking at DoReflection, I am not sure why it's non-const. Is it because `DoDecay` changes the member variables of the anharmonicDecay class which in turn changes the BoundaryProcess member?

kelseymh: If `DoDecay()` is calling into non-const functions of `anharmonicDecay`, then it needs to be non-const itself, so that it has access to a non-const pointer.

nttenpas: That makes sense, thank you! That looks like it's exactly what's happening. So `DoDecay` is non-const, and so `GetReflectedVector` can be non-const to set the `stepSize` member var.

kelseymh: Oh, ugh.  So `G4CMPAnharmonicDecay::DoDecay()` is non-const only because it's manipulating the `output` data member for debugging, and similarity with `MakeTTSecondaries()` and `MakeLTSecondaries()` -- it's only about the use of `output`.

nttenpas: I also see that `DoDecay` sets the other data members: `fBeta`, `fGamma`, `fLambda`, and `fMu`. That make sense why everything here has to be non-const.

PR #42, opened 2025-04-15
G4CMP-469

cstonewhitehead:
I have added the Silica files to feature branch G4CMP-469 and believe they are ready to be merged. I followed the same convention as Israel and included the citations as comments in the config files.


kelseymh:
Could you add the two new materials to the table of materials in README.md?

cstonewhitehead:
Should be added now. Let me know if anything isn't as anticipated

PR #43, opened 2025-04-23
Merge 468 to 317

nttenpas:
G4CMP-468 handles the volume assignment for phonons after a displaced reflection. This branch improved the stepping algorithm to avoid any escaping phonons and implemented tests to avoid stuck jobs in the event a phonon does escape.

### DIFF
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -7,7 +7,9 @@ Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
 *** ADD NEW TAG MESSAGES BELOW G4CMP-317 LINE, UPDATE DATE OF G4CMP-317 ***
-2025-03-25  G4CMP-317 : Implement surface 'displacement' for phonon reflection.
+2025-04-23  G4CMP-317 : Implement surface 'displacement' for phonon reflection.
+2025-04-23  G4CMP-468 : Create custom phonon particle change to assign correct
+                        volume after a displacement reflection.
 2025-03-25  G4CMP-463 : Add surface displacement step size/limit macro commands.
 2025-02-06  G4CMP-459 : Support reflection displacement search at hard corners.
 2025-01-30  G4CMP-453 : Add rotations and utilities for PhononVelocityIsInward.
--- a/library/src/G4CMPBoundaryUtils.cc
+++ b/library/src/G4CMPBoundaryUtils.cc
@@ -257,7 +257,7 @@ G4bool G4CMPBoundaryUtils::CheckStepBoundary(const G4Step& aStep,
 	     << " NEW SURFACE POINT: " << surfacePoint << G4endl;
     }
 
-    postIn = preSolid->Inside(postPos);
+    postIn = preSolid->Inside(surfacePoint);
     if (buVerboseLevel>2) {
       G4cout << "\n Is adjusted location on surface of preStep Volume? "
 	     << (postIn==kOutside ? "outside" :
--- a/library/src/G4CMPBoundaryUtils.cc
+++ b/library/src/G4CMPBoundaryUtils.cc
@@ -246,7 +246,14 @@ G4bool G4CMPBoundaryUtils::CheckStepBoundary(const G4Step& aStep,
   // If post-step position not proper surface point, compute intersection
   if (postIn != kSurface) {
     G4ThreeVector along = (postPos-prePos).unit();	// Trajectory direction
+    G4cout << "OLD SURFACE POINT: " << surfacePoint << G4endl;
     surfacePoint = prePos + preSolid->DistanceToOut(prePos,along)*along;
+    G4cout << "NEW SURFACE POINT: " << surfacePoint << G4endl;
+    postIn = preSolid->Inside(postPos);
+    G4cout << "\n Is adjusted postStep location on surface of preStep Volume? "
+	   << (postIn==kOutside ? "outside" :
+	       postIn==kInside  ? "inside" :
+	       postIn==kSurface ? "surface" : "INVALID") << G4endl;
 
     // Double check calculation -- point "must" now be on surface!
     if (preSolid->Inside(surfacePoint) != kSurface) {
--- a/library/src/G4CMPBoundaryUtils.cc
+++ b/library/src/G4CMPBoundaryUtils.cc
@@ -26,6 +26,7 @@
 // 20211207  Replace G4Logical*Surface with G4CMP-specific versions.
 // 20250413  Protect debugging messages with verbosity.
 // 20250415  Suppress error for same PV if starting at boundary.
+// 20250423  Remove error suppression for starting at boundary.
 
 #include "G4CMPBoundaryUtils.hh"
 #include "G4CMPConfigManager.hh"
@@ -104,7 +105,7 @@ G4bool G4CMPBoundaryUtils::GetBoundingVolumes(const G4Step& aStep) {
   prePV = aStep.GetPreStepPoint()->GetPhysicalVolume();
   postPV = aStep.GetPostStepPoint()->GetPhysicalVolume();
 
-  if (prePV == postPV && aStep.GetPreStepPoint()->GetStepStatus() != fGeomBoundary) {
+  if (prePV == postPV) {
     if (buVerboseLevel) {
       G4cerr << procName << " ERROR: fGeomBoundary status set, but"
 	     << " pre- and post-step volumes are identical!" << G4endl;
--- a/library/src/G4CMPBoundaryUtils.cc
+++ b/library/src/G4CMPBoundaryUtils.cc
@@ -246,15 +246,24 @@ G4bool G4CMPBoundaryUtils::CheckStepBoundary(const G4Step& aStep,
 
   // If post-step position not proper surface point, compute intersection
   if (postIn != kSurface) {
+    if (buVerboseLevel>2)
+      G4cout << " OLD SURFACE POINT: " << surfacePoint << G4endl;
+
     G4ThreeVector along = (postPos-prePos).unit();	// Trajectory direction
-    G4cout << "OLD SURFACE POINT: " << surfacePoint << G4endl;
     surfacePoint = prePos + preSolid->DistanceToOut(prePos,along)*along;
-    G4cout << "NEW SURFACE POINT: " << surfacePoint << G4endl;
+    if (buVerboseLevel>2) {
+      G4cout << " moving preStep by " << preSolid->DistanceToOut(prePos,along)
+	     << " along " << along << G4endl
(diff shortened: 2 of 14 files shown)


library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -213,61 +225,53 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
   }
 
   // Update trackInfo wavevector and particleChange's group velocity and momentum direction
   // reflectedKDir is in global coordinates here - no conversion needed
   FillParticleChange(particleChange, aTrack, reflectedKDir);
-  G4ThreeVector vdir = *particleChange.GetMomentumDirection();
+  const G4ThreeVector vdir = *particleChange.GetMomentumDirection();
+
+  // If displacement occured: update the particle change and navigator for volume assignment
+  if (refltype == "specular" && *particleChange.GetPosition() != surfacePoint) {
```
kelseymh, line 249: Would like to use an enum for this test.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -213,61 +225,53 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
+    phParticleChange.ProposePosition(surfacePoint);
+    phParticleChange.ProposeTouchableHandle(aStep.GetPreStepPoint()->GetTouchableHandle());
```
kelseymh, line 251: Can we move this to a additional implementation of FillParticleChange?

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -213,61 +225,53 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
+    aStep.GetPostStepPoint()->SetStepStatus(fPostStepDoItProc);
+
+    G4Navigator* navigator = G4TransportationManager::GetTransportationManager()->GetNavigatorForTracking();
+    navigator->LocateGlobalPointWithinVolume(surfacePoint);
+    G4double safety = aStep.GetPostStepPoint()->GetSafety();
+    navigator->ComputeStep(surfacePoint, vdir, aStep.GetStepLength(), safety);
```
kelseymh, line 257: Put this into a wrapper function.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -294,20 +299,23 @@ GetReflectedVector(const G4ThreeVector& waveVector,
...
-  G4double kTanMag = kTan.mag();
+  G4ThreeVector axis = kPerpV.cross(kTan).unit();
+  G4double phi = 0.;
   EInside isIn = solid->Inside(stepLocalPos);
 
   G4int nAttempts = 0;
 
   // debugging only DELETE
   G4ThreeVector oldkTan = kTan;
   G4ThreeVector oldkPerpV = kPerpV;
```
kelseymh, line 317: Should this be removed?

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -336,8 +344,8 @@ GetReflectedVector(const G4ThreeVector& waveVector,
 
   // Assumes everything is in Global. Just add the GetGlobal in the loop conditions.
   while (!G4CMP::PhononVelocityIsInward(theLattice, mode,
-   GetGlobalDirection(reflectedKDir), GetGlobalDirection(newNorm))
-	 && nAttempts++ < nStepLimit) {
+   GetGlobalDirection(reflectedKDir), GetGlobalDirection(newNorm),
+   GetGlobalPosition(stepLocalPos)) && nAttempts++ < nStepLimit) {
     // Save previous loop values
     oldstepLocalPos = stepLocalPos;
     oldNorm = newNorm;
```
kelseymh, line 351: Are the "old" variables still being used for anything?

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -406,13 +419,14 @@ GetReflectedVector(const G4ThreeVector& waveVector,
...
     G4cout << (GetProcessName()+"::GetReflectedVector").c_str()
       << ": Phonon displacement failed after " << nAttempts - 1 
       << " attempts. Doing diffuse reflection at surface point: " << surfacePoint << G4endl;
 
     // reflectedKDir and stepLocalPos will be in global coordinates
-    reflectedKDir = GetLambertianVector(surfNorm, mode);
+    reflectedKDir = GetLambertianVector(surfNorm, mode, surfacePoint);
     stepLocalPos = surfacePoint;
   }
   else{
```
kelseymh, line 432: Missing space

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -439,15 +453,16 @@ GetReflectedVector(const G4ThreeVector& waveVector,
...
   G4int nTries = 0;
   do {
     reflectedKDir = G4CMP::LambertReflection(surfNorm);
   } while (nTries++ < maxTries &&
 	   !G4CMP::PhononVelocityIsInward(theLattice, mode,
-					  reflectedKDir, surfNorm));
+					  reflectedKDir, surfNorm, surfPoint));
 
   return reflectedKDir;
 }
```
kelseymh, line 468: Could we move this whole function out of the boundary process?  Maybe to `G4CMPUtils`, with `PhononVelocityIsInward()`?   If we do that, then PhononElectrode could just call it, rather than having to have a copy of the implementation.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -554,10 +569,8 @@ AdjustToClosestSurfacePoint(const G4VSolid* solid,
   if (isIn == kSurface) return;
 
   // Angles to be adjusted in place by OptimizeSurfaceAdjustAngle
-  // Best initial conditions will be the surface normal
-  G4ThreeVector surfNorm = solid->SurfaceNormal(stepLocalPos);
-  G4double bestTheta = surfNorm.theta();
-  G4double bestPhi = surfNorm.phi();
+  G4double bestTheta = pi / 2;
+  G4double bestPhi = 0;
```
kelseymh, line 573: Why not use the surface normal (or its reflection) as the initial guess?

library/src/G4CMPPhononElectrode.cc
```diff
@@ -145,7 +146,8 @@ ProcessAbsorption(const G4Track& track, const G4Step& step, G4double EDep,
     do {
       reflectedKDir = G4CMP::LambertReflection(surfNorm);
     } while (!G4CMP::PhononVelocityIsInward(theLattice, pol,
-                                            kmag*reflectedKDir, surfNorm));
+                                            kmag*reflectedKDir, surfNorm,
+                                            track.GetPosition()));
```
kelseymh, line 150: See my comment about moving this to G4CMPUtils rather than duplicating the code from BoundaryProcess.

kelseymh (changes requested):

ChangeHistory
```diff
@@ -7,7 +7,9 @@ Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
 *** ADD NEW TAG MESSAGES BELOW G4CMP-317 LINE, UPDATE DATE OF G4CMP-317 ***
-2025-03-25  G4CMP-317 : Implement surface 'displacement' for phonon reflection.
+2025-04-23  G4CMP-317 : Implement surface 'displacement' for phonon reflection.
+2025-04-23  G4CMP-468 : Create custom phonon particle change to assign correct
```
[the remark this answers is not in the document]
nttenpas: This is today's date

library/src/G4CMPPhononBoundaryProcess.cc
nttenpas: We need to keep these two. They are used for the edge reflection code.

library/src/G4CMPPhononBoundaryProcess.cc
nttenpas: Maybe the request is desynced? Are you talking about the space between `mode,` and `surfacePoint`?

ChangeHistory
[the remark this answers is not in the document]
kelseymh: I see it.  My mistake.

library/src/G4CMPPhononBoundaryProcess.cc
kelseymh: Lines 417-418 should be consolidated to
```
  } else {
```

library/src/G4CMPPhononBoundaryProcess.cc
kelseymh: Oh!  It's a gimbal lock situation.  I understand.  It might be worth putting in a one-line comment mentioning why you have it start at the theta pole.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -213,61 +225,53 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
+    G4ThreeVector vec2 = GetLambertianVector(surfNorm, mode, surfacePoint);
 
     sec1->SetMomentumDirection(vec1);
     sec2->SetMomentumDirection(vec2);
 
     return;
   } else if (random < downconversionProb + specProb) {
     reflectedKDir = GetReflectedVector(waveVector, surfNorm, mode, surfacePoint); // Modify surfacePoint & surfNorm in place
-    //particleChange.ProposePosition(surfacePoint);
     refltype = "specular";
```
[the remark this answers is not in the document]
nttenpas: I agree. I am thinking about this. I included the refltype in the test because the volume assignment was wonky and CheckStepBoundary can change the position. With that resolved, I think the test should just be:
`*particleChange.GetPosition() != surfacePoint`

Can you think of a situation where we would adjust to the surface and not adjust the touchable?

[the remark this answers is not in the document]
kelseymh: In principle, the only time the postStep point should be moving at all is when we successfully apply the displacement loop.  In that case, there shouldn't be any change to the position or the touchable.

library/src/G4CMPBoundaryUtils.cc
```diff
@@ -102,7 +104,7 @@ G4bool G4CMPBoundaryUtils::GetBoundingVolumes(const G4Step& aStep) {
   prePV = aStep.GetPreStepPoint()->GetPhysicalVolume();
   postPV = aStep.GetPostStepPoint()->GetPhysicalVolume();
 
-  if (prePV == postPV) {
+  if (prePV == postPV && aStep.GetPreStepPoint()->GetStepStatus() != fGeomBoundary) {
```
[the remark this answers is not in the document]
nttenpas: That was added before you checked for the 0 step length. I removed this line.

PR #44, opened 2025-04-24
Merge 465 to 317

nttenpas:
Initial implementation of the G4CMPSolidUtils class. This class contains the custom solid functions used for the edge reflection code in PhononBoundaryProcess.

### DIFF
--- a/library/include/G4CMPSolidUtils.hh
+++ b/library/include/G4CMPSolidUtils.hh
@@ -24,18 +24,26 @@
 #include "globals.hh"
 #include "G4ThreeVector.hh"
 
+class G4AffineTransform;
+class G4RotationMatrix;
 class G4VSolid;
 class G4VTouchable;
 
 class G4CMPSolidUtils {
   public:
     // Default constructor
-    G4CMPSolidUtils() : theSolid(0), theTouchable(0), verboseLevel(0),
+    G4CMPSolidUtils() : theSolid(0), theTransform(0), verboseLevel(0),
                         verboseLabel("G4CMPSolidUtils") {;}
 
-    // Direct constructor with solid, touchable, and verbose
-    G4CMPSolidUtils(const G4VSolid* solid, const G4VTouchable* touch,
+    // Direct constructor with solid & transform for client code in local frame
+    G4CMPSolidUtils(const G4VSolid* solid, G4int verbose, G4String vLabel);
+    G4CMPSolidUtils(const G4VSolid* solid, const G4AffineTransform& trans,
                     G4int verbose, G4String vLabel);
+    G4CMPSolidUtils(const G4VSolid* solid, const G4RotationMatrix& rot,
+                    const G4ThreeVector& disp, G4int verbose, G4String vLabel);
+
+    // Direct constructor with touchable for client code in global frame
+    G4CMPSolidUtils(const G4VTouchable* touch, G4int verbose, G4String vLabel);
 
     // Copy operation
     G4CMPSolidUtils& operator=(const G4CMPSolidUtils& right);
@@ -45,8 +53,14 @@ class G4CMPSolidUtils {
       theSolid = solid;
     }
 
-    void SetTouchable(const G4VTouchable* touch) {
-      theTouchable = touch;
+    void SetTransform(const G4AffineTransform& trans) {
+      theTransform = trans;
+    }
+
+    void SetTransform(const G4RotationMatrix& rot,
+                      const G4ThreeVector& disp) {
+      theTransform.SetNetRotation(rot);
+      theTransform.SetNetTranslation(disp);
     }
 
     void SetVerboseLevel(G4int verbose) {
@@ -63,7 +77,8 @@ class G4CMPSolidUtils {
     }
 
     const G4VSolid* GetSolid() const { return theSolid; }
-    const G4VTouchable* GetTouchable() const { return theTouchable; }
+    const G4AffineTransform GetTransform() const { return theTransform_LtoG; }
+    const G4AffineTransform GetInverseTransform() const { return theTransform_GtoL; }
     G4int GetVerboseLevel() const { return verboseLevel; }
     G4String GetVerboseLabel() const {return verboseLabel; }
 
@@ -116,9 +131,18 @@ class G4CMPSolidUtils {
                             const G4ThreeVector& localPos,
                             G4ThreeVector& surfNorm) const;
 
+  protected:
+      void TransformLocalToGlobal(G4ThreeVector& point, G4ThreeVector& dir) const;
+      void TransformGlobalToLocal(G4ThreeVector& point, G4ThreeVector& dir) const;
+      void TransformToGlobalPoint(G4ThreeVector& point) const;
+      void TransformToLocalPoint(G4ThreeVector& point) const;
+      void TransformToGlobalDirection(G4ThreeVector& dir) const;
+      void TransformToLocalDirection(G4ThreeVector& dir) const;
+
   private:
     const G4VSolid* theSolid;
-    const G4VTouchable* theTouchable;
+    const G4AffineTransform theTransform_LtoG;
+    const G4AffineTransform theTransform_GtoL;
     G4int verboseLevel;
     G4String verboseLabel;
 };
--- a/library/include/G4CMPSolidUtils.hh
+++ b/library/include/G4CMPSolidUtils.hh
@@ -31,7 +31,7 @@ class G4CMPSolidUtils {
   public:
     // Default constructor
     G4CMPSolidUtils() : theSolid(0), theTouchable(0), verboseLevel(0),
-                        verboseLabel("") {;}
+                        verboseLabel("G4CMPSolidUtils") {;}
 
     // Direct constructor with solid, touchable, and verbose
     G4CMPSolidUtils(const G4VSolid* solid, const G4VTouchable* touch,
@@ -55,6 +55,10 @@ class G4CMPSolidUtils {
 
     void SetVerboseLevel(G4int verbose, G4String vLabel) {
       verboseLevel = verbose;
+      verboseLabel = vLabel;
+    }
+
+    void SetVerboseLabel(G4String vLabel) {
       verboseLabel = vLabel;
     }
 
--- a/library/include/G4CMPSolidUtils.hh
+++ b/library/include/G4CMPSolidUtils.hh
@@ -130,7 +130,7 @@ class G4CMPSolidUtils {
                             const G4ThreeVector& localPos,
                             G4ThreeVector& surfNorm) const;
 
-  protected:
+    // Internal transformations
     void TransformLocalToGlobal(G4ThreeVector& point, G4ThreeVector& dir) const;
     void TransformGlobalToLocal(G4ThreeVector& point, G4ThreeVector& dir) const;
     void TransformToGlobalPoint(G4ThreeVector& point) const;
--- a/library/include/G4CMPSolidUtils.hh
+++ b/library/include/G4CMPSolidUtils.hh
@@ -8,11 +8,15 @@
 /// directly included in the G4VSolid class. Some example of utilites are:
 /// 1) Adjusting a position to the nearest surface point without the surface normal.
 /// 2) Adjusting a particle to the nearest edge along the surface.
+/// 3) Reflecting a tangent vector against a "hard" edge.
 ///
-/// The user will need to create this class directly with a G4VSolid and/or
-/// touchable, or can initialize these objects later on.
+/// The user will need to create this class directly with a G4VSolid (and transform) or
+/// a touchable. If the transform or touchable is not provided, an indentity transform is
+/// used.
 ///
-/// Note: All coordinates/vectors passed into functions must be in the solid's local coordinate system.
+/// Note: All coordinates/vectors passed into functions must be in the global coordinate system.
+/// If a global-to-local transform is supplied, the input coordinates and vectors, and any output
+/// results, will be in the coordinate system of that transform.
 //
 // $Id$
 //
@@ -36,14 +40,18 @@ class G4CMPSolidUtils {
                         verboseLabel("G4CMPSolidUtils") {;}
 
     // Direct constructor with solid & transform for client code in local frame
-    G4CMPSolidUtils(const G4VSolid* solid, G4int verbose, G4String vLabel);
+    G4CMPSolidUtils(const G4VSolid* solid, G4int verbose=0, const G4String& vLabel="G4CMPSolidUtils");
+
     G4CMPSolidUtils(const G4VSolid* solid, const G4AffineTransform& trans,
-                    G4int verbose, G4String vLabel);
+                    G4int verbose=0, const G4String& vLabel="G4CMPSolidUtils");
+
     G4CMPSolidUtils(const G4VSolid* solid, const G4RotationMatrix& rot,
-                    const G4ThreeVector& disp, G4int verbose, G4String vLabel);
+                    const G4ThreeVector& disp, G4int verbose=0,
+                    const G4String& vLabel="G4CMPSolidUtils");
 
     // Direct constructor with touchable for client code in global frame
-    G4CMPSolidUtils(const G4VTouchable* touch, G4int verbose, G4String vLabel);
+    G4CMPSolidUtils(const G4VTouchable* touch, G4int verbose=0,
+                    const G4String& vLabel="G4CMPSolidUtils");
 
     // Copy operation
     G4CMPSolidUtils& operator=(const G4CMPSolidUtils& right);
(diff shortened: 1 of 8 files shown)


library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,59 @@
+/***********************************************************************\
+ * This software is licensed under the terms of the GNU General Public *
+ * License version 3 or later. See G4CMP/LICENSE for the full license. *
+\***********************************************************************/
+
+/// \file library/include/G4CMPSolidUtils.hh
+/// \brief Definition of the G4CMPSolidUtils class
```
kelseymh, line 7: Ideally, you would describe what G4CMPSolidUtils is used for, or how it works, in a couple of sentences.  This line is meant as the "template placeholder", but I have been really horrible about doing this properly myself.

library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,59 @@
...
+
+  // Default constructor
+  G4CMPSolidUtils() : theSolid(0), verboseLevel(0) {;}
+
+  void Initialize(const G4VSolid* solid);
+
+  void SetVerboseLevel(G4int verbose=0) { verboseLevel = verbose; }
+
+  const G4VSolid* GetSolid() { return theSolid; }
+
```
kelseymh, line 32: Be sure to add a comment, as well as text up in the header comment, that the position coordinate _must_ be passed in the solid's local coordinate system (but see my comment to the constructor).

library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,59 @@
...
+#include "globals.hh"
+#include "G4ThreeVector.hh"
+
+class G4VSolid;
+
+class G4CMPSolidUtils {
+  public:
+
+  // Default constructor
+  G4CMPSolidUtils() : theSolid(0), verboseLevel(0) {;}
```
kelseymh, line 25: It would be useful to have a non-default constructor where you pass in the G4VSolid* directly.

It might also be useful to have another constructor (and Set function) where you can pass in the global-to-local transform (available from the Touchable) that you want applied to coordinates and vectors.

That second constructor would mean that the user could pass in global coordinates to the member functions and it would translate to/from local internally.

library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,59 @@
...
+                            G4ThreeVector& stepLocalPos,
+                            G4double high, const G4int curvedSurf) const;
+
+  // Modifies kTan and newNorm in place
+  void ReflectAgainstEdge(G4ThreeVector& kTan,
+                          const G4ThreeVector& stepLocalPos,
+                          G4ThreeVector& newNorm) const;
+  protected:
+    G4CMPSolidUtils(const G4VSolid* solid);
+    G4CMPSolidUtils& operator=(const G4CMPSolidUtils& right);
```
kelseymh, line 52: I don't see a reason to forbid these.  Move them up to the public section.  You have an implementation in the .cc file already, which is great.

library/src/G4CMPSolidUtils.cc
```diff
@@ -0,0 +1,245 @@
+/***********************************************************************\
+ * This software is licensed under the terms of the GNU General Public *
+ * License version 3 or later. See G4CMP/LICENSE for the full license. *
+\***********************************************************************/
+
+/// \file library/src/G4PhononReflection.cc
+/// \brief Implementation of the G4PhononReflection class
+//
+// This utility class handles general G4VSolid processes.
```
kelseymh, line 9: Be more specific here, as I mentioned in the .hh file.  For multiple lines, you need to use the "///" notation for Doxygen to pick up the whole thing.  As currently written, line 9 would not get used by Doxygen.

library/include/G4CMPSolidUtils.hh
kelseymh, line 32: Are there any other functions we use with `G4VSolid` directly which could be moved into this class.  Anything in `G4CMPGeometryUtils`?

nttenpas: Do we want to move the simple DistanceToIn, Inside, etc. here? Happy to, just wasn't sure because it would be the same code.

library/include/G4CMPSolidUtils.hh
nttenpas: Do you think we should require global coordinates instead? I can add the touchable as a member var, and directly use the G4CMPGeometryUtils RotateTo* and GetLocal*.

That would be simple, and would align with the other utility classes. But this is a solid class, so the local frame is going to be used everywhere. If people use this class, I would guess that they are working in local frames already.

I can see it both ways. I am not sure which is best.

library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,59 @@
...
+
+class G4VSolid;
+
+class G4CMPSolidUtils {
+  public:
+
+  // Default constructor
+  G4CMPSolidUtils() : theSolid(0), verboseLevel(0) {;}
+
+  void Initialize(const G4VSolid* solid);
```
nttenpas, line 27: @kelseymh this works, but I am a bit confused why. The solid member variable is const, so why can I set it again with this function? In the boundary code I think it follows:
1) blank class initialization - no solid
2) when the reflection code starts, we initialize the solid with the track's current solid

Will this break when we have multiple solids? If so, should I be deleting the whole class and making a new one using the constructor that takes in the solid argument?

library/src/G4CMPSolidUtils.cc
```diff
@@ -0,0 +1,245 @@
...
+    // Find the normal for the reflection "surface"
+    refNorm = (edgeVec.cross(newNorm)).unit();
+    if (refNorm * kTan < 0) refNorm *= -1;
+
+    // Reflect kTan against reflection "surface"
+    kTan -= 2*((kTan * refNorm) * refNorm);
+  }
+
+  if (verboseLevel>3) {
+    G4cout << "G4CMPSolidUtils::ReflectAgainstEdge"
```
nttenpas, line 236: New thought: Should I use the processName from the class that is using these utilities? That might make debugging easier to tell where these functions got called from.

library/include/G4CMPSolidUtils.hh
nttenpas: I created functions:
    G4double GetDistanceToSolid(const G4ThreVector& localPos);
    G4double GetDistanceToSolid(const G4ThreVector& localPos,
                                const G4ThreeVector& dir);
                                
which handle both DistanceToIn and DistanceToOut, so the user doesn't have to determine which one to use. I still am unsure about the Inside function.

library/include/G4CMPSolidUtils.hh
nttenpas: Or I could re-write the functions to return new points instead of adjusting them in place.

kelseymh: Take a look at what we do in CDMSSolidUtils.  This might be worth a conversation, because there is no obvious answer.

library/include/G4CMPSolidUtils.hh
kelseymh: I wouldn't bother with replicating call-throughs into G4VSolid.  The client code must already have the solid in order to instantiate G4CMPSolidUtils.  The idea is to wrap up non-trivial stuff ({{GetDistance}} above is a good example).  For this example, I could even imagine a version of {{GetDistance()}} that takes a _non-const_ vector as input, and returns the direction to the point along with the distance.

library/include/G4CMPSolidUtils.hh
kelseymh: You wrote above,

> If people use this class, I would guess that they are working in local frames already

I'm thinking just the opposite!  By encapsulating it, you could let the client code stick with global coordinates everywhere (we could even consider using this within `IsVelocityInward()`!) if they either register a transformation (`G4Transform3D` or `G4AffineTransform`), or pass in a Touchable.  That's more along the lines of what Geant4 wants to do.

library/include/G4CMPSolidUtils.hh
kelseymh: This works fine the way it is, but I'm not a big fan.  That default constructor means that when you create the object, it isn't actually usable (you'd get a segfault if you tried calling some of the utility functions).  The client code _must_ call `Initialize()` in order for the object to be usable.  So my preference would be for the constructor to take a `G4VSolid*` (you could have the argument default to null to keep this behaviour, if you want it), and rename `Initialize()` to `SetSolid()`.

If you're willing to change BoundaryProcess, I think you don't really need to have a persistent data member that you reuse.  You could just have one of these objects created in `PostStepDoIt()` (or maybe just in `DoReflection()`), and let it get deleted at the end of the function.

kelseymh (changes requested):

library/src/G4CMPSolidUtils.cc
```diff
@@ -0,0 +1,334 @@
...
+
+    if (isIn == kSurface) low = mid; // Move out
+    else high = mid; // Move in
+  }
+
+  if (verboseLevel>3) {
+    G4double remDist = (theSolid->DistanceToIn(localPos) >=
+                        theSolid->DistanceToOut(localPos))
+                      ? theSolid->DistanceToIn(localPos)
+                      : theSolid->DistanceToOut(localPos);
```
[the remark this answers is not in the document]
nttenpas: This one is actually slightly different. I wrote this output a long time ago to verify the edge adjustment worked, and we are always, at most, within 1pm of the solid. I think this should be removed - thank you for pointing this out. 

I implemented the GetDistanceToSurface functions on the last push. They are essentially wrappers so I didn't have to do:

```if kInside use DistanceToOut
else if kOutside use DistanceToIn
else 0```

I am aware we could sum the two distances, because one _should_ be 0. But I prefer the explicit method considering all the issues I've seen with the geometry so far :-D

File: library/src/G4CMPSolidUtils.cc (hunk 7 above)
Comment (nttenpas) [reply library/src/G4CMPSolidUtils.cc:325]: Understood! Thank you

File: library/src/G4CMPSolidUtils.cc (hunk 8 above)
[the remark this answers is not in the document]
Comment (nttenpas) [reply library/src/G4CMPSolidUtils.cc:271]: Well... I lied. This apparently was my entire problem with the new class not doing the same as the 317 branch. As it turns out, I need to do:

```
if kInside use DistanceToOut
else use DistanceToIn
```

I think the reasoning is that the DistanceToIn will include that tolerance, which is required by all the other functions. When I used the original function, I almost tripled the failed reflections.

File: library/include/G4CMPSolidUtils.hh [hunk 9]
```diff
@@ -0,0 +1,102 @@
...
+    G4CMPSolidUtils(const G4VSolid* solid);
+
+    // Copy operation
+    G4CMPSolidUtils& operator=(const G4CMPSolidUtils& right);
+
+    // Set and get member variables
+    void Initialize(const G4VSolid* solid);
+    void Initialize(const G4VSolid* solid, G4int verbose);
+    void SetVerboseLevel(G4int verbose=0) { verboseLevel = verbose; }
+
```
kelseymh, line 44: You asked in the .cc file about having a useful name string for debugging output.  Maybe you want to add an additional "name" or "label" string data member, which could default to being "G4CMPSolidUtils", but which the client code could set with a `SetLabel()` function.

kelseymh (changes requested):

library/include/G4CMPSolidUtils.hh
kelseymh: In addition to its own mutator, you could add a second optional argument to SetVerboseLevel, passing in a desired "label" string that way.

library/include/G4CMPSolidUtils.hh
```diff
@@ -25,25 +25,43 @@
...
 
 class G4VSolid;
+class G4VTouchable;
 
 class G4CMPSolidUtils {
   public:
     // Default constructor
-    G4CMPSolidUtils() : theSolid(0), verboseLevel(0) {;}
+    G4CMPSolidUtils() : theSolid(0), theTouchable(0), verboseLevel(0),
+                        verboseLabel("") {;}
```
kelseymh, line 34: Do you want to make the default label "G4CMPSolidUtils" in this constructor, just for consistency?

library/src/G4CMPSolidUtils.cc
```diff
@@ -262,8 +253,8 @@ AdjustToEdgePosition(const G4ThreeVector& vTan,
     else high = mid; // Move in
   }
 
-  if (verboseLevel>3) {
-    G4cout << "G4CMPSolidUtils::AdjustToEdgePosition"
+  if (verboseLevel > 1) {
```
kelseymh, line 256: Not suggesting a change, but why did you move this from verbosity 4 down to verbosity 2?

nttenpas: I lowered this and the `ReflectAgainstEdge` down because the output isn't too overwhelming. I was also thinking that these are fairly sensitive, so it could be useful to get this debugging output without a bunch of other output. Do you have thoughts on what it should be?

On a related note: On the latest push, I lowered `AdjustToClosestSurfacePoint` to verbose 1 to match the other instance of that function.

library/include/G4CMPSolidUtils.hh
nttenpas: That should already be in there (sort of). I have two `SetVerboseLevel` classes. One that takes the verbose level and one that takes a level and the label. I didn't have a `SetVerboseLabel` for the label only. I added that on the latest push.

library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,172 @@
...
+/// \file library/include/G4CMPSolidUtils.hh
+/// \brief This class contains custom utilities for G4VSolids that are not
+/// directly included in the G4VSolid class. Some example of utilites are:
+/// 1) Adjusting a position to the nearest surface point without the surface normal.
+/// 2) Adjusting a particle to the nearest edge along the surface.
+///
+/// The user will need to create this class directly with a G4VSolid and/or
+/// touchable, or can initialize these objects later on.
+///
+/// Note: All coordinates/vectors passed into functions must be in the solid's local coordinate system.
```
kelseymh, line 15: You might append to this, "If a global-to-local transform is supplied, the input coordinates and vectors, and any output results, will be in the coordinate system of that transform."

library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,172 @@
...
+class G4VTouchable;
+
+class G4CMPSolidUtils {
+  public:
+    // Default constructor
+    G4CMPSolidUtils() : theSolid(0), theTransform(G4AffineTransform()), verboseLevel(0),
+                        verboseLabel("G4CMPSolidUtils") {;}
+
+    // Direct constructor with solid & transform for client code in local frame
+    G4CMPSolidUtils(const G4VSolid* solid, G4int verbose, G4String vLabel);
```
kelseymh, line 39: I'd suggest providing default values for `verbose` and `vLabel` in all these constructors.  That way the client code doesn't _have_ to pass in zeroes for those.

library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,172 @@
...
+
+    void SetTransform(const G4AffineTransform& trans) {
+      theTransform = trans;
+    }
+
+    void SetTransform(const G4RotationMatrix& rot,
+                      const G4ThreeVector& disp) {
+      theTransform.SetNetRotation(rot);
+      theTransform.SetNetTranslation(disp);
+    }
```
kelseymh, line 64: You may want to add a third version of `SetTransform()` which takes a rotation pointer (and of course tests if it's null before dereferencing).

library/src/G4CMPSolidUtils.cc
```diff
@@ -0,0 +1,392 @@
...
+#include "G4UnitsTable.hh"
+
+
+// Direct constructors
+
+G4CMPSolidUtils::G4CMPSolidUtils(const G4VSolid* solid=0,
+                                 G4int verbose=0,
+                                 G4String vLabel="G4CMPSolidUtils")
+  : theSolid(solid),
+    theTransform(G4AffineTransform(G4RotationMatrix::IDENTITY, G4ThreeVector(0,0,0))),
```
kelseymh, line 38: You don't need to do this explicitly.  `G4AffineTransform()` by itself returns the identity.

library/include/G4CMPSolidUtils.hh
kelseymh, line 39: I didn't catch this earlier, sorry!  Use `const G4String& vLabel` for the constructor and Set function argument.  Otherwise, you end up creating and deleting a temporary.

library/src/G4CMPSolidUtils.cc
```diff
@@ -0,0 +1,392 @@
...
+#include "G4VSolid.hh"
+#include "G4VTouchable.hh"
+#include "G4UnitsTable.hh"
+
+
+// Direct constructors
+
+G4CMPSolidUtils::G4CMPSolidUtils(const G4VSolid* solid=0,
+                                 G4int verbose=0,
+                                 G4String vLabel="G4CMPSolidUtils")
```
kelseymh, line 36: As I noted in the .hh file, use `const G4String& vLabel` for this argument, everywhere.

kelseymh (changes requested):

library/include/G4CMPSolidUtils.hh
```diff
@@ -0,0 +1,186 @@
...
+
+    void SetTransform(const G4RotationMatrix* rot,
+                      const G4ThreeVector& disp) {
+      if (rot) SetTransform(*rot, disp);
+      else SetTransform(G4RotationMatrix::IDENTITY, disp);
```
kelseymh, line 77: You could be extra clever, and do this with a ternary operator (but you certainly don't have to!):
```
SetTransform((rot?*rot:G4RotationMatrix::IDENTITY), disp);
```

nttenpas: I am always willing to be extra clever!

PR #47, opened 2025-05-01
461 to 317

nttenpas:
This includes the new version of the flat skipper that uses the edge reflection code, as well as an additional utility for finding the distance to a solid's bounding box.

### DIFF
--- a/library/src/G4CMPSolidUtils.cc
+++ b/library/src/G4CMPSolidUtils.cc
@@ -21,6 +21,8 @@
 // $Id$
 //
 // 20250424  G4CMP-465 -- Create G4CMPSolidUtils class.
+// 20250429  G4CMP-461 -- Add function for skipping detector flats.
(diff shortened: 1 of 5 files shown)


library/src/G4CMPSolidUtils.cc
```diff
@@ -261,7 +263,8 @@ void G4CMPSolidUtils::AdjustToClosestSurfacePoint(G4ThreeVector& pos,
 // Get edge position along vTan from pos
 G4ThreeVector G4CMPSolidUtils::GetEdgePosition(const G4ThreeVector& vTan,
                                                const G4ThreeVector& pos,
-                                               G4double high, const G4int curvedSurf) const {
+                                               G4double high,
+                                               const G4int curvedSurf) const {
```
kelseymh, line 267: I'm curious about why you broke this into two lines -- did it go past 80 columns?

Also, it looks like line 266 might have extra spaces at the end of the line; those should be removed.

PR #50, opened 2025-05-09
Merge 317 onto develop

nttenpas:
This includes the work done to model surface mode displacements for specular reflections on the curved surfaces of a detector.

### DIFF
--- a/library/src/G4CMPPhononBoundaryProcess.cc
+++ b/library/src/G4CMPPhononBoundaryProcess.cc
@@ -30,6 +30,7 @@
 // 20220910  G4CMP-299 -- Use fabs(k) in absorption test.
 // 20240718  G4CMP-317 -- Initial implementation of surface displacement.
 // 20250124  G4CMP-447 -- Use FillParticleChange() to update wavevector and Vg.
+// 20250204  G4CMP-459 -- Support reflection displacement search at hard corners.
 
 #include "G4CMPPhononBoundaryProcess.hh"
 #include "G4CMPAnharmonicDecay.hh"
@@ -245,7 +246,7 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
 	   << "\n from " << surfacePoint << "\n   to " << stepPos << G4endl;
 
     G4ThreeVector stepLocal = GetLocalPosition(stepPos);
-    G4VSolid* solid = aStep.GetPreStepPoint()->GetPhysicalVolume()->GetLogicalVolume()->GetSolid();
+    G4VSolid* solid = GetCurrentVolume()->GetLogicalVolume()->GetSolid();
 
     EInside place = solid->Inside(stepLocal);
     G4cout << " After trial step, " << (place==kInside ? "inside"
@@ -327,7 +328,14 @@ GetReflectedVector(const G4ThreeVector& waveVector,
 
     // FIXME: Find point on surface nearest to stepLocalPos, and reset
     surfAdjust = solid->DistanceToIn(stepLocalPos, -newNorm);	// Get distance along normal from new position back to detector surface
-    stepLocalPos -= surfAdjust * newNorm;			// Adjust position to be back on detector surface
+
+    // Adjust position to be back on detector surface
+    // If surfAdjust is large we have stepped off a hard corner and need correction
+    if (surfAdjust > 1.*mm) {
+      AdjustPositionAtEdge(stepLocalPos, kTan, stepSize); // Modifies stepLocalPos and kTan in place
+    } else {
+      stepLocalPos -= surfAdjust * newNorm;
+    }
 
     // Get rotation axis perpendicular to waveVector-normal plane
     axis = kPerpV.cross(kTan).unit();
@@ -347,7 +355,7 @@ GetReflectedVector(const G4ThreeVector& waveVector,
     reflectedKDir = kTan + kPerpV;
 
     // Debugging: Can be removed?
-    G4ThreeVector vDir = theLattice->MapKtoVDir(mode, GetGlobalDirection(reflectedKDir));
+    G4ThreeVector vDir = theLattice->MapKtoVDir(mode, reflectedKDir);
 
     // FIXME: Need defined units
     if (verboseLevel>3) {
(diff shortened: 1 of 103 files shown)


library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -192,128 +227,249 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
+	   << "\n New momentum direction " << vdir << G4endl;
   }
 
   // If reflection failed, report problem and kill the track
-  if (!G4CMP::PhononVelocityIsInward(theLattice,mode,reflectedKDir,surfNorm)) {
+  if (!G4CMP::PhononVelocityIsInward(theLattice,mode,reflectedKDir,surfNorm, surfacePoint)) {
     G4Exception((GetProcessName()+"::DoReflection").c_str(), "Boundary010",
-		JustWarning, ("Phonon "+refltype+" reflection failed").c_str());
-    DoSimpleKill(aTrack, aStep, aParticleChange);
+		JustWarning, ("Phonon "+refltype+" reflection failed"+"\nPhonon mode at time of death: "+G4PhononPolarization::Label(mode)).c_str());
```
kelseymh, line 268: The "\n" can be in the middle of the string literal, which might be better depending on how a given compiler does grouping.  And maybe break it out to a temporary variable, to allow line breaks:
```
G4String msg = "Phonon "+refltype+" reflection failed";
msg += "\nPhonon mode at time of death: " + G4PhononPolarization::Label(mode);
```
Then you can have `msg.c_str()` in the `G4Exception()` call.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -192,128 +227,249 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
+  G4ThreeVector newNorm = surfNorm;
+  RotateToLocalDirection(newNorm);
   
-  if (verboseLevel>2) {
-    G4cout << " specular reflection with normal " << surfNorm
-	   << "\n Perpendicular wavevector " << kPerp*surfNorm
-	   << " (mag " << kPerp << ")" << G4endl;
+  // Initialize solid object and utilities
+  G4VSolid* solid = GetCurrentVolume()->GetLogicalVolume()->GetSolid();
+  G4CMPSolidUtils* solidUtils = new G4CMPSolidUtils(solid, verboseLevel,
```
kelseymh, line 300: This does not need to be a pointer.  If you make it just a local variable, i.e.,
```
G4CMPSolidUtils solidUtils(solid, verboseLevel, GetProcessName());
```
then you don't have to `delete` it later.  It'll just go away when the function returns.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -192,128 +227,249 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
-GetLambertianVector(const G4ThreeVector& surfNorm, G4int mode) const {
-  G4ThreeVector reflectedKDir;
-  const G4int maxTries = 1000;
-  G4int nTries = 0;
-  do {
-    reflectedKDir = G4CMP::LambertReflection(surfNorm);
-  } while (nTries++ < maxTries &&
-	   !G4CMP::PhononVelocityIsInward(theLattice, mode,
-					  reflectedKDir, surfNorm));
+  delete solidUtils;
```
kelseymh, line 460: If you make `solidUtils` a simple local variable, you don't need to this delete.  Also, if the function breaks/returns in the middle for some reason, it'll also be cleaned up automatically.

library/src/G4CMPProcessUtils.cc
```diff
@@ -75,6 +81,7 @@
 #include "G4PhysicalConstants.hh"
 #include "G4RotationMatrix.hh"
 #include "G4SystemOfUnits.hh"
+#include "G4Step.hh"
```
kelseymh, line 84: Move up one line (alphabetical order).

library/src/G4CMPProcessUtils.cc
```diff
@@ -132,8 +139,9 @@ void G4CMPProcessUtils::LoadDataForTrack(const G4Track* track) {
 
     const G4ParticleDefinition* pd = track->GetParticleDefinition();
     G4Track* tmp_track = const_cast<G4Track*>(track);
-    tmp_track->SetMomentumDirection(
-      theLattice->MapKtoVDir(G4PhononPolarization::Get(pd), tempvec));
+    const G4ThreeVector localVDir = theLattice->MapKtoVDir(G4PhononPolarization::Get(pd),
```
kelseymh, line 142: I don't want to use a per-call temporary here.  We have so many of these calls in the job that it could make a time difference.  

Since you're already using tempvec for the k vector, I think maybe we should add a`tempvec2` data member.  It's ugly, but will improve performance (so that should be part of the comment associated with it).

kelseymh (changes requested):

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -192,128 +227,249 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
+    G4cout << "New surface position " << *particleChange.GetPosition()/mm 
+	   << " mm\n New wavevector direction " << reflectedKDir
+	   << "\n New momentum direction " << vdir << G4endl;
   }
 
   // If reflection failed, report problem and kill the track
-  if (!G4CMP::PhononVelocityIsInward(theLattice,mode,reflectedKDir,surfNorm)) {
+  if (!G4CMP::PhononVelocityIsInward(theLattice,mode,reflectedKDir,surfNorm, surfacePoint)) {
+    G4String msg = "Phonon " + refltype + " reflection failed";
+    msg += "\nPhonon mode at time of death: " + *G4PhononPolarization::Label(mode);
```
kelseymh, line 268: I didn't realize the `Label()` returned a C-style `const char*`, for which there is no operator+().  That's probably why it worked before, because of the grouping of multiple operator+().  Maybe you want to just use the full phonon name and make it one line:
```
  G4String msg = G4PhononPolarization::Name(mode) + " " + refltype + " reflection failed";
```

kelseymh (changes requested):

kelseymh (approved):

PR #51, opened 2025-05-10
Merge 483 to 317

nttenpas:
This ensures backwards compatibility for the vector utilities that were created during the 468 work.

### DIFF
--- a/library/src/G4CMPPhononBoundaryProcess.cc
+++ b/library/src/G4CMPPhononBoundaryProcess.cc
@@ -429,7 +429,7 @@ GetSpecularVector(const G4ThreeVector& waveVector,
   RotateToGlobalPosition(stepLocalPos);
 
   if (!G4CMP::PhononVelocityIsInward(theLattice, mode, reflectedKDir, newNorm,
-                                     GetGlobalPosition(stepLocalPos))) {
+                                     stepLocalPos)) {
     if (verboseLevel) {
       G4cerr << GetProcessName() << "::GetSpecularVector"
 	     << ": Phonon displacement failed after " << nAttempts - 1
--- a/library/src/G4CMPUtils.cc
+++ b/library/src/G4CMPUtils.cc
@@ -20,7 +20,7 @@
 // 20250130  G4CMP-453 -- Apply coordinate rotations in PhononVelocityIsInward
 // 20250422  G4CMP-468 -- Add displaced point test to PhononVelocityIsInward.
 // 20250423  G4CMP-468 -- Add function to get diffuse reflection vector.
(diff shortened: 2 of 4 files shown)


library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -429,7 +429,7 @@ GetSpecularVector(const G4ThreeVector& waveVector,
   RotateToGlobalPosition(stepLocalPos);
 
   if (!G4CMP::PhononVelocityIsInward(theLattice, mode, reflectedKDir, newNorm,
-                                     GetGlobalPosition(stepLocalPos))) {
+                                     stepLocalPos)) {
```
kelseymh, line 432: Good catch!  I missed that double rotation, because of the name.

library/src/G4CMPUtils.cc
```diff
@@ -211,6 +212,13 @@ void G4CMP::FillHit(const G4Step* step, G4CMPElectrodeHit* hit) {
 // Generate cos(theta) law for diffuse reflection, ensuring that computed
 // vector is directed inward with respect to the surface normal.
 
+G4ThreeVector
+G4CMP::GetLambertianVector(const G4LatticePhysical* theLattice,
+			   const G4ThreeVector& surfNorm, G4int mode) {
+  const G4ThreeVector surfPoint = GetCurrentTrack()->GetPosition();
+  return GetLambertianVector(theLattice, surfNorm, mode, surfPoint);
```
kelseymh, line 219: You could do this as a one-liner if you wanted to:
```
  return GetLambertianVector(theLattice, surfNorm, mode, GetCurrentTrack()->GetPosition());
```
But the way you have it would let you print the vector in the debugger, so it doesn't need to be changed.

library/src/G4CMPUtils.cc
```diff
@@ -241,6 +249,14 @@ G4ThreeVector G4CMP::LambertReflection(const G4ThreeVector& surfNorm) {
 // Check that phonon is properly directed from the volume surface
 // waveVector and surfNorm need to be in global coordinates
 
+G4bool G4CMP::PhononVelocityIsInward(const G4LatticePhysical* lattice,
+                                     G4int mode,
+                                     const G4ThreeVector& waveVector,
+                                     const G4ThreeVector& surfNorm) {
+  const G4ThreeVector surfacePos = GetCurrentTrack()->GetPosition();
+  return PhononVelocityIsInward(lattice, mode, waveVector, surfNorm, surfacePos);
```
kelseymh, line 257: Same note as above.  I don't think you need to change anything.

kelseymh (approved):
Just observations.  I think this can be merged as it is.  Thank you!

PR #54, opened 2025-07-28
G4cmp-490 to G4CMP-404

emichaud4:
I have fix the phonon energy being calculated in the HV frame in G4CMPLukeScattering.cc instead of in the lattice frame : a simple rotation out of the HV frame for electrons. I also added a small security checks to make sure the phonon energy is below the electron energy. If that is the case, do an another roll.

Ready to be merged to G4CMP-404!

### DIFF
--- a/library/src/G4CMPLukeScattering.cc
+++ b/library/src/G4CMPLukeScattering.cc
@@ -235,8 +235,9 @@ G4VParticleChange* G4CMPLukeScattering::PostStepDoIt(const G4Track& aTrack,
     
     // Get recoil wavevector (in HV frame), convert to new local momentum
     k_recoil = ktrk - qvec;
-    if (IsElectron()) qvec =
-      lat->SphericalToEllipsoidalTranformation(iValley, qvec);
+    if (IsElectron()) {
+      qvec = lat->SphericalToEllipsoidalTranformation(iValley, qvec);
+    }
     qmag = qvec.mag();
     Ephonon = MakePhononEnergy(qmag);
     // Make sure energy is conserved


library/src/G4CMPLukeScattering.cc
```diff
@@ -235,17 +235,26 @@ G4VParticleChange* G4CMPLukeScattering::PostStepDoIt(const G4Track& aTrack,
     
     // Get recoil wavevector (in HV frame), convert to new local momentum
     k_recoil = ktrk - qvec;
+    if (IsElectron()) {qvec = lat->SphericalToEllipsoidalTranformation(iValley, qvec);}
     qmag = qvec.mag();
```
kelseymh, line 239: For a single line if-statement, please don't use braces.

If this line doesn't fit into 80 columns, please move the `qvec = ...` part to the line immediately below the if, indented by two spaces, without braces.

kelseymh (changes requested):

library/src/G4CMPLukeScattering.cc
emichaud4: Removed the braces and moved part of the equation below the if to respect the 80 spaces!

kelseymh: Nope.  Please put the _whole_ equation on the second line.  Don't split the if-statement that way.  If the long expression doesn't fit, then split it and use braces.

kelseymh (changes requested):

kelseymh (approved):

library/src/G4CMPLukeScattering.cc
emichaud4: I used braces and put the qvec = ... whole equation directly below the if, with 2 spaces

PR #56, opened 2025-08-26
G4 cmp 496

wlamberson1:
There is a problem with the phonon wavevector and momentum direction coming out of the DoDecay code in G4CMPPhononBoundaryProcess.

Also, when the Lambertian Reflection is used to reassign the wavevector, the original mode is being used for both secondaries. Each secondary mode should be used instead.

We can use the secondary polarization in the LambertianVector call and then use the FillParticleChange method added in 317 to update the wavevector and group velocity assignment directly in the new daughter tracks.


relineha:
Noticing a few merge conflicts. Currently reviewing.

relineha:
@wlamberson1 Do we need the two RotateToGlobalDirection calls? (In other words, are the Lambertian Vectors returned in the lines above those not in the global coordinate system already? Looking in G4CMPGeometryUtils::GetSurfaceNormal(), I think the surface norm is returned in the global coordinate system.

kelseymh:
> @wlamberson1 Do we need the two RotateToGlobalDirection calls? (In other words, are the Lambertian Vectors returned in the lines above those not in the global coordinate system already? Looking in G4CMPGeometryUtils::GetSurfaceNormal(), I think the surface norm is returned in the global coordinate system.

Oh, yuck.  You are quite right that `G4CMPGeometryUtils::GetSurfaceNormal()` is explicitly rotating to global.  That's what a G4Navigator version would do, but it's still probably not consistently assumed everywhere.

wlamberson1:
> @kelseymh As far as I can tell with the code I've touched for phonon boundary interactions, I think it is scoped correctly in most places, but there may be more nefarious assumptions hiding that I may have missed.
> 
> In any case, if Wade says it's okay, I'll remove those two calls to the RotateToGlobalDirection, and with that I think I have all I need to reserve the merge conflicts. (There is some duplication of code on my end that Wade nicely packaged into a G4CMPUtils function that I want to remove in favor of that function, but that should be about the rest of it.)

Sounds good to me.

wlamberson1:
I went ahead and rectified both merge conflicts and removed the two RotateToGlobalDirection() lines. Once all reviews are approved I'll merge.

library/src/G4CMPPhononBoundaryProcess.cc
```diff
@@ -328,14 +329,17 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
       G4cout << "DR Function Point CA | after DoDecay in surface-mediated "
 	     << "downconversion" << G4endl;
     }
+  
+    G4int mode1 = G4PhononPolarization::Get(sec1->GetParticleDefinition());
+    G4int mode2 = G4PhononPolarization::Get(sec2->GetParticleDefinition());
```
kelseymh, line 334: On these two lines, the code Ryan introduced is more compact, using a nice function already defined, so I think you can keep it, viz.:
```G4int mode1 = GetPolarization(sec1);```

File: library/src/G4CMPPhononBoundaryProcess.cc [hunk 2]
```diff
@@ -328,14 +329,17 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
-    G4int mode2 = GetPolarization(sec2);
+    G4ThreeVector vec1 = G4CMP::GetLambertianVector(theLattice, surfNorm, mode1,
+                                                    surfacePoint);
+    G4ThreeVector vec2 = G4CMP::GetLambertianVector(theLattice, surfNorm, mode2,
+                                                    surfacePoint);
+
+    UpdatePhononWavevector(*sec1, vec1);
+    UpdatePhononWavevector(*sec2, vec2);
 
     //Debugging
```
kelseymh, line 344: I think a lot of this big block of code following this this point should be removed.  The calls to `GetLambertianVector()` already includes a loop to ensure that the generated wavevector is suitable for inward propagation.  The additional cross-checks shouldn't be required.  

@relineha Would you be willing to check out this branch (G4CMP-496) and make that change if appropriate?

kelseymh (changes requested):

library/src/G4CMPPhononBoundaryProcess.cc
relineha: Agreed with this.

library/src/G4CMPPhononBoundaryProcess.cc
relineha: Sounds great -- thanks Wade!

relineha (approved):
As far as I can tell, duplicated code is now gone and the fix is properly implemented.

PR #57, opened 2025-08-29
Protect G4CMPProcessUtils::FillParticleChange with Phonon Check.

wlamberson1:
While reviewing [G4CMP-496](https://jira.slac.stanford.edu/browse/G4CMP-496), we realized the G4CMPProcessUtils::FillParticleChange, which uses phonon specific functions to get the group velocity (and velocity magnitude), does not protect against someone calling this function for electrons. 

We should protect the phonon-specific functions with a "IsPhonon()" if-statement. Return immediately if the particle is a charge.


wlamberson1:
@relineha This is a two liner, and is the last thing I have to merge into feature/TrackedFilmResponse.

PR #63, opened 2025-09-29
G4CMP 478

jyamaoka:
Lowered the minimum energy for secondaries produced by G4CMPDriftTrapIonization from 1e-3 to 1e-6.  Geant4 requires some energy but need to preserve energy/momentum conservation.

Before
Before
After
After


kelseymh (changes requested):
Please update the top-level ChangeHistory. 

Just out of curiosity, why did you merge develop onto this branch?  It should have been recent enough to not have any conflicts anywhere.

jyamaoka:
I guess habit.  I'll try to pay more attention to see if it is really necessary in the future.

PR #70, opened 2026-01-19
G4CMP-513

wlamberson1:
Split DriftBoundaryProcess::DoElectronReflection() and DoHoleReflection() into separate specular and diffuse functions. May be possible to have simple top-level DoReflection() function which handles the specular/diffuse split, and a similar unified diffuse reflection function (just throw outgoing transport momentum direction).

Specular reflection needs to be different for electrons, because of the mass tensor. We reflect the wavevector, convert back to transport momentum, and fall back to diffuse scattering if the resulting momentum is not inward. This will be a modification of the specular electron reflection code currently on branch [G4CMP-503](https://jira.slac.stanford.edu/browse/G4CMP-503), which uses velocity instead of wavevector.

### DIFF
--- a/library/include/G4CMPDriftBoundaryProcess.hh
+++ b/library/include/G4CMPDriftBoundaryProcess.hh
@@ -30,6 +30,7 @@
 // 20251013  Add functions for specular and diffuse electron reflection.
 // 20251204  G4CMP-511 -- Create parallel Lambertian reflection code for charges.
 // 20251210  G4CMP-518 -- Make PhononVelocityIsInward() generic.
+// 20250219  G4CMP-513 : Provide separate specular and diffuse reflection for charges.
 
 #ifndef G4CMPDriftBoundaryProcess_h
 #define G4CMPDriftBoundaryProcess_h 1
@@ -64,18 +65,11 @@ protected:
   virtual void DoReflection(const G4Track& aTrack,const G4Step& aStep,
 			    G4ParticleChange& aParticleChange);
 
-  virtual void DoReflectionElectron(const G4Track& aTrack,const G4Step& aStep,
-				    G4ParticleChange& aParticleChange);
+  virtual G4ThreeVector DoSpecularReflection(const G4Track& aTrack,const G4Step& aStep);
 
-  virtual G4ThreeVector DoSpecularElectron(const G4ThreeVector& inDir,
-					   const G4ThreeVector& surfNorm,
-					   const G4ThreeVector& surfPos) const;
+  virtual G4ThreeVector DoSpecRefElectron(const G4Track& aTrack,const G4Step& aStep);
 
-  virtual G4ThreeVector DoDiffuseElectron(const G4ThreeVector& surfNorm,
-					  const G4ThreeVector& surfPos) const;
-
-  virtual void DoReflectionHole(const G4Track& aTrack,const G4Step& aStep,
-				G4ParticleChange& aParticleChange);
+  virtual G4ThreeVector DoSpecRefHole(const G4Step& aStep);
 
   // Called when maximum bounces have been recorded; does recombination
   virtual void DoFinalReflection(const G4Track& aTrack,const G4Step& aStep,
--- a/library/include/G4CMPDriftBoundaryProcess.hh
+++ b/library/include/G4CMPDriftBoundaryProcess.hh
@@ -67,9 +67,9 @@ protected:
 
   virtual G4ThreeVector DoSpecularReflection(const G4Track& aTrack,const G4Step& aStep);
 
-  virtual G4ThreeVector DoSpecRefElectron(const G4Track& aTrack,const G4Step& aStep);
+  virtual G4ThreeVector DoSpecularElectron(const G4Track& aTrack,const G4Step& aStep);
 
-  virtual G4ThreeVector DoSpecRefHole(const G4Step& aStep);
+  virtual G4ThreeVector DoSpecularHole(const G4Step& aStep);
 
   // Called when maximum bounces have been recorded; does recombination
   virtual void DoFinalReflection(const G4Track& aTrack,const G4Step& aStep,
--- a/library/include/G4CMPDriftBoundaryProcess.hh
+++ b/library/include/G4CMPDriftBoundaryProcess.hh
@@ -62,14 +62,14 @@ protected:
   virtual void DoAbsorption(const G4Track& aTrack, const G4Step& aStep,
 			    G4ParticleChange& aParticleChange);
 
-  virtual void DoReflection(const G4Track& aTrack,const G4Step& aStep,
+  virtual void DoReflection(const G4Track& aTrack, const G4Step& aStep,
 			    G4ParticleChange& aParticleChange);
 
-  virtual G4ThreeVector DoSpecularReflection(const G4Track& aTrack,const G4Step& aStep);
+  virtual G4ThreeVector DoSpecularReflection(const G4Track& aTrack, const G4Step& aStep);
 
-  virtual G4ThreeVector DoSpecularElectron(const G4Track& aTrack,const G4Step& aStep);
+  virtual G4ThreeVector DoSpecularElectron(const G4Track& aTrack, const G4Step& aStep);
 
-  virtual G4ThreeVector DoSpecularHole(const G4Step& aStep);
+  virtual G4ThreeVector DoSpecularHole(const G4Track& aTrack, const G4Step& aStep);
 
   // Called when maximum bounces have been recorded; does recombination
   virtual void DoFinalReflection(const G4Track& aTrack,const G4Step& aStep,
--- a/library/src/G4CMPDriftBoundaryProcess.cc
+++ b/library/src/G4CMPDriftBoundaryProcess.cc
@@ -181,12 +181,12 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
   G4ThreeVector reflDir;
 
   if (random < specProb) {
-    reflDir = DoSpecularReflection(aTrack, aStep);
+    reflP = DoSpecularReflection(aTrack, aStep);
   } else { // Do diffuse reflection (electrons & holes)
-    reflDir = LambertianReflection(theLattice, G4CMP::GetSurfaceNormal(aStep), GetCurrentValley());
+    reflP = LambertianReflection(theLattice, G4CMP::GetSurfaceNormal(aStep), GetCurrentValley());
   }
 
-  FillParticleChange(GetCurrentValley(), aTrack.GetKineticEnergy(), reflDir);
+  FillParticleChange(GetCurrentValley(), reflP);
 }
 
 G4ThreeVector G4CMPDriftBoundaryProcess::
@@ -194,20 +194,20 @@ DoSpecularReflection(const G4Track& aTrack, const G4Step& aStep) {
   if (verboseLevel>1)
     G4cout << GetProcessName() << ": Electron reflected" << G4endl;
 
-  G4ThreeVector reflDir;
+  G4ThreeVector reflP;
   
-  if (IsElectron())  reflDir = DoSpecRefElectron(aTrack, aStep);
-  else if (IsHole()) reflDir = DoSpecRefHole(aStep);
+  if (IsElectron())  reflP = DoSpecularElectron(aTrack, aStep);
+  else if (IsHole()) reflP = DoSpecularHole(aStep);
   else {
     G4Exception("G4CMPDriftBoundaryProcess::DoReflection", "Boundary004",
                 EventMustBeAborted, "Invalid particle for this process.");
   }
 
-  return reflDir;
+  return reflP;
 }
 
 G4ThreeVector G4CMPDriftBoundaryProcess::
-DoSpecRefElectron(const G4Track& aTrack, const G4Step& aStep) {
+DoSpecularElectron(const G4Track& aTrack, const G4Step& aStep) {
   G4ThreeVector k = GetLocalWaveVector(aTrack);
   G4ThreeVector surfNorm = G4CMP::GetSurfaceNormal(aStep);
 
@@ -223,6 +223,12 @@ DoSpecRefElectron(const G4Track& aTrack, const G4Step& aStep) {
   // Specular reflection reverses wavevector along normal
   G4double dirNorm = k * surfNorm;
   G4ThreeVector k -= 2.*dirNorm*surfNorm;
+
+  // If reflected velocity is outward facing, fall back to diffuse reflection
+  G4int mode = GetPolarization(aStep.GetTrack());
+  if (!G4CMP::VelocityIsInward(theLattice, mode, k, surfNorm)) {
+    k = LambertianReflection(theLattice, G4CMP::GetSurfaceNormal(aStep), GetCurrentValley());
+  }
   
   if (verboseLevel>2) {
     G4cout << " New wavevector direction " << k.unit() << G4endl;
@@ -243,11 +249,11 @@ DoSpecRefElectron(const G4Track& aTrack, const G4Step& aStep) {
     G4cout << " Cross-check new wavevector direction " << vnew.unit() << G4endl;
   }
 
-  return k;
+  return p;
 }
 
 G4ThreeVector G4CMPDriftBoundaryProcess::
-DoSpecRefHole(const G4Step& aStep) {
+DoSpecularHole(const G4Step& aStep) {
   if (verboseLevel>1) {
     G4cout << GetProcessName() << "DoSpecularHole " << G4endl;
   }
@@ -266,7 +272,7 @@ DoSpecRefHole(const G4Step& aStep) {
     G4cout << " New momentum direction " << momDir << G4endl;
   }
   
-  return momDir;
+  return aTrack.GetMomentum().mag() * momDir;
 }
 
 
--- a/library/src/G4CMPDriftBoundaryProcess.cc
+++ b/library/src/G4CMPDriftBoundaryProcess.cc
@@ -228,7 +228,7 @@ DoSpecularElectron(const G4Track& aTrack, const G4Step& aStep) {
   // If reflected velocity is outward facing, fall back to diffuse reflection
   G4int mode = GetPolarization(aStep.GetTrack());
   if (!G4CMP::VelocityIsInward(theLattice, mode, k, surfNorm)) {
-    k = LambertianReflection(theLattice, surfNorm, GetCurrentValley());
+    return LambertianReflection(theLattice, surfNorm, GetCurrentValley());
   }
   
   if (verboseLevel>2) {
--- a/library/src/G4CMPDriftBoundaryProcess.cc
+++ b/library/src/G4CMPDriftBoundaryProcess.cc
@@ -19,6 +19,7 @@
 // 20250927  AbsorbTrack() should use '&&' to require that both conditions pass
 // 20251204  G4CMP-511 -- Create parallel Lambertian reflection code for charges.
 // 20251210  G4CMP-518 -- Make PhononVelocityIsInward() generic.
+// 20250219  G4CMP-513 : Provide separate specular and diffuse reflection for charges.
 
 #include "G4CMPDriftBoundaryProcess.hh"
 #include "G4CMPConfigManager.hh"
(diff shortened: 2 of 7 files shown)


library/include/G4CMPDriftBoundaryProcess.hh
```diff
@@ -64,18 +65,11 @@ protected:
...
 			    G4ParticleChange& aParticleChange);
 
-  virtual void DoReflectionElectron(const G4Track& aTrack,const G4Step& aStep,
-				    G4ParticleChange& aParticleChange);
+  virtual G4ThreeVector DoSpecularReflection(const G4Track& aTrack,const G4Step& aStep);
 
-  virtual G4ThreeVector DoSpecularElectron(const G4ThreeVector& inDir,
-					   const G4ThreeVector& surfNorm,
-					   const G4ThreeVector& surfPos) const;
+  virtual G4ThreeVector DoSpecRefElectron(const G4Track& aTrack,const G4Step& aStep);
```
kelseymh: I think I'd keep the original function names ("DoSpecularElectron", etc.).  Also, I see that you've changed the function signature so you're starting over from the track and step in each of the functions.  Why not extract the incident direction and surface normal once, pass those into the functions that do the computations, and then return the outgoing direction (or full outgoing vector)?  If you need access to the track (e.g., for the electron valley) you can use the G4CMPProcessUtils::GetCurrentTrack() function.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -167,69 +168,69 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
+    G4cout << " New wavevector direction " << k.unit() << G4endl;
+  }
   
-  // Convert velocity back to momentum and update direction
-  RotateToLocalDirection(outDir);
-  G4ThreeVector p = theLattice->MapV_elToP(GetCurrentValley(), outDir);
+  // Convert wavevector back to momentum and update direction
+  RotateToGlobalDirection(k);
+  G4ThreeVector p = theLattice->MapV_elToP(GetCurrentValley(), k);
   RotateToGlobalDirection(p);
```
kelseymh: This must be tested for "is inward" before you return anything. If the is-inward test fails, then you should fall back to calling diffuse reflection (this is another reason to use the same function signature for both).

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -239,45 +240,36 @@ DoSpecularElectron(const G4ThreeVector& inDir,
     G4ThreeVector vnew = theLattice->MapPtoV_el(GetCurrentValley(),
 						GetLocalDirection(p));
     RotateToGlobalDirection(vnew);
-    G4cout << " Cross-check new v dir  " << vnew.unit() << G4endl;
+    G4cout << " Cross-check new wavevector direction " << vnew.unit() << G4endl;
   }
 
-  return outDir;
+  return k;
```
kelseymh: Is this the correct thing to return?  Is this what the top-level DoReflection() expects to receive, and to pass into FillParticleChange()?

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -239,45 +240,36 @@ DoSpecularElectron(const G4ThreeVector& inDir,
...
   G4double momNorm = momDir * surfNorm;
   momDir -= 2.*momNorm*surfNorm;
   
-  if (verboseLevel>2)
+  if (verboseLevel>2) {
     G4cout << " New momentum direction " << momDir << G4endl;
+  }
   
-  aParticleChange.ProposeMomentumDirection(momDir);
+  return momDir;
```
kelseymh: This is a unit vector, right?  Is this the correct thing to be returning to the parent function for use in FillParticleChange?

kelseymh (changes requested):

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: I need to multiply by the magnitude, don't I?

kelseymh: That's the question, isn't it?  Do you multiply by the magntude _here_ (and then have to make sure you do that same thing in the other functions), or do you return a unit vector _here_, and do the multiplication and call to FillParticleChange up in the higher-level DoReflection() function?  I'm not giving you an answer, I'm asking what you think we should do (and why).

library/include/G4CMPDriftBoundaryProcess.hh
wlamberson1: The reason I passed aStep and aTrack was because DoSpecRefHole() uses it to get momDir.

So, if I caclulated surfNorm and passed it to both functions I'd also have to calculated momDir and pass it to DoSpecRefHole(). For aTrack, I'd have to calculate k and pass it to DoSpecRefElectron.

If I did all that, it would make DoSpecularReflection() messy, and everything would still be called the same number of times. I figure having the surfNorm line twice (but still only calling it once) it a better tradeoff than how messy DoSpecularReflection() would look.

wlamberson1: It would add five lines to DoSpecularReflection()

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: Thank you!

library/include/G4CMPDriftBoundaryProcess.hh
kelseymh: I think I understand, but I think I'm missing something.  Maybe you could put on a slide your comparison of the two situations?

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: I am confused about this. My code looks identical to the reflection in G4CMPPhononBoundaryProcess. Doesn't FillParticleChange take the reflectedKDir? So it can be a unit vector?

kelseymh: There are separate, different FillParticleChange() functions for charges vs. phonons, so you need to check the right one to see what it expects and what it does with it.  See, for example, G4CMP-580, which is what triggered Emile's comment to you in G4CMP-513.

wlamberson1: Okay, when I look at the .hh and .cc containing the two FillParticleChange() functions it is not clear if either is different for phonons/charges.

wlamberson1: I found the correct FillParticleChange() functions

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: I am looking at the Geant$ documentation. I'm not seeing a convenient way to get the magnitude. I do see that I can propose the new momentum direction, and then I think I can use GetMomentum() to get the new momDir*magnitude.

kelseymh: Yes, that's mostly right.  Use `aTrack.GetMomentum().mag()` to get the magnitude, then you can pass in a "full momentum" to FillParticleChange().  But if you are calling our own `FillParticleChange()` (which is what you need to always do for electrons), then you should not call `particleChange.ProposeMomentumDirection()` directly.  each `FillParticleChange()` function will take care of _all_ the necessary `ProposeXYZ()` calls.  

For consistency, you can also call `FillParticleChange()` for holes (in which case the valley index would be -1 or ignored), which should allow you do so some function unification.

library/include/G4CMPDriftBoundaryProcess.hh
wlamberson1: Maybe it's more clear with this latest push. Especially for DoSpecularElectron. If not I can explain further!
Actually, with the check for outward facing velocity, it would actually make even less sense to change what's passed in.

library/src/G4CMPDriftBoundaryProcess.cc
kelseymh: Ah, I see where you got messed up, and it's our fault for inconsistency.  The phonon `FillParticleChange()` functions got moved to `G4CMPProcessUtils` long ago, but the charge-track `FillParticleCharge()` are member functions of `G4CMPVDriftProcess`.   I'll need to check what Ryan did for the QPs (or if he didn't do anything).

library/include/G4CMPDriftBoundaryProcess.hh
```diff
@@ -64,18 +65,11 @@ protected:
   virtual void DoReflection(const G4Track& aTrack,const G4Step& aStep,
 			    G4ParticleChange& aParticleChange);
 
-  virtual void DoReflectionElectron(const G4Track& aTrack,const G4Step& aStep,
-				    G4ParticleChange& aParticleChange);
+  virtual G4ThreeVector DoSpecularReflection(const G4Track& aTrack,const G4Step& aStep);
```
kelseymh: Shouldn't there be a high-level `DoDiffuseReflection(aTrack,aStep)` function here, called from `DoReflection()`?  Even if all it does it call through to `LambertReflection()`, that would make the parallelism of the structure, and the ability in future to re-split things, more obvious.

library/include/G4CMPDriftBoundaryProcess.hh
```diff
@@ -64,18 +65,11 @@ protected:
...
-					   const G4ThreeVector& surfNorm,
-					   const G4ThreeVector& surfPos) const;
+  virtual G4ThreeVector DoSpecularElectron(const G4Track& aTrack,const G4Step& aStep);
 
-  virtual G4ThreeVector DoDiffuseElectron(const G4ThreeVector& surfNorm,
-					  const G4ThreeVector& surfPos) const;
-
-  virtual void DoReflectionHole(const G4Track& aTrack,const G4Step& aStep,
-				G4ParticleChange& aParticleChange);
+  virtual G4ThreeVector DoSpecularHole(const G4Step& aStep);
```
kelseymh: Why doesn't this take the same `G4Track&` argument as the other functions?  Having a different signature will be confusing to the next person in the future who wants to deal with this (even if we had documentation, inconsistent signatures make the code confusing).

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -167,69 +168,75 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
-  // Get outward normal from current volume
-  G4ThreeVector surfNorm = G4CMP::GetSurfaceNormal(aStep);
+  G4double random = G4UniformRand();
 
-  // Check whether step has proper boundary-stopped geometry
-  G4ThreeVector surfacePoint;
-  if (!CheckStepBoundary(aStep, surfacePoint)) {
-    if (verboseLevel>2)
-      G4cout << " Boundary point moved to " << surfacePoint << G4endl;
+  G4ThreeVector reflDir;
```
kelseymh: Is this correct?  The lines below have a variable named `reflP`, which I don't see defined.  If this variable isn't used, you should have gotten a compiler warning when you built.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -239,45 +246,36 @@ DoSpecularElectron(const G4ThreeVector& inDir,
...
 }
 
-
-void G4CMPDriftBoundaryProcess::
-DoReflectionHole(const G4Track& /*aTrack*/, const G4Step& aStep,
-		 G4ParticleChange& /*aParticleChange*/) {
-  if (verboseLevel>1)
-    G4cout << GetProcessName() << ": Hole reflected" << G4endl;
+G4ThreeVector G4CMPDriftBoundaryProcess::
+DoSpecularHole(const G4Step& aStep) {
```
kelseymh: See my comment in the .hh file for consistent signature.  In this code, you would specify the argument as "`const G4Track& /*aTrack*/`") (with the variable name commented out), to both avoid a compiler warning about unused argument, and to indicate to future developers that the argument just happens not to be used in the current implementation.

kelseymh (changes requested):

library/include/G4CMPDriftBoundaryProcess.hh
wlamberson1: I can add that if you'd like! But this seems to match what is done in G4CMPPhononBoundaryProcess.cc and I did leave a comment there to make it clear it's diffuse reflection for electrons and holes.

library/include/G4CMPDriftBoundaryProcess.hh
wlamberson1: I can add that parameter, but it won't be used.

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: Oops! Thank you.

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: Would it also be commented out in the .hh file?

wlamberson1: Also, should I switch the order in both functions so the commented out one is last?

kelseymh: > Would it also be commented out in the .hh file?

In principle, it doesn't matter; the variable _name_ is irrelevant in the declaration (.hh file).  We don't normally leave out the names, so please stick with our convention.

> should I switch the order in both functions so the commented out one is last?

No, please do not.  All the other functions have the same order, and that order matches what Geant4 uses for PostStepDoIt().  Consistency is important for future understanding.

library/include/G4CMPDriftBoundaryProcess.hh
kelseymh: That's exactly what I'm asking for.  Consistency helps with understanding.

library/include/G4CMPDriftBoundaryProcess.hh
kelseymh: Oh, I see.  `G4CMPPhononBoundaryProcess` calls `DoLambertReflection()` directly; in effect, it's using "Lambert" instead of "Diffuse", eh?  In that case, you can keep this the same way for consistency.  Thank you for explaining!

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -167,69 +168,75 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
-    G4cout << " New velocity direction " << outDir.unit() << G4endl;
+  if (verboseLevel>2) {
+    G4cout << " New wavevector direction " << k.unit() << G4endl;
+  }
   
-  // Convert velocity back to momentum and update direction
-  RotateToLocalDirection(outDir);
-  G4ThreeVector p = theLattice->MapV_elToP(GetCurrentValley(), outDir);
+  // Convert wavevector back to momentum and update direction
+  RotateToGlobalDirection(k);
```
kelseymh: I don't think this is correct.  The Map functions should be assuming the input vectors are in the _volume local_ frame (the functions will do the necessary lattice and valley rotations internally).

If k comes out of `LambertianReflection()` in global coordinates (which happens if the surface normal is global), then you should be doing `RotateToLocalDirection(k)` here.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -167,69 +168,75 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
-  G4double dirNorm = inDir * surfNorm;
-  G4ThreeVector outDir = inDir - 2.*dirNorm*surfNorm;
+  // Specular reflection reverses wavevector along normal
+  G4double dirNorm = k * surfNorm;
+  k -= 2.*dirNorm*surfNorm;
+
+  // If reflected velocity is outward facing, fall back to diffuse reflection
+  G4int mode = GetPolarization(aStep.GetTrack());
+  if (!G4CMP::VelocityIsInward(theLattice, mode, k, surfNorm)) {
+    k = LambertianReflection(theLattice, G4CMP::GetSurfaceNormal(aStep), GetCurrentValley());
```
kelseymh: Why are you recomputing `GetSurfaceNormal()` here?  You have the `surfNorm` value already, you should just reuse that for consistency.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -167,69 +168,75 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
 }
 
 G4ThreeVector G4CMPDriftBoundaryProcess::
-DoSpecularElectron(const G4ThreeVector& inDir,
-		   const G4ThreeVector& surfNorm,
-		   const G4ThreeVector& /*surfPos*/) const {
-  if (verboseLevel>2) G4cout << " DoSpecularElectron " << surfNorm << G4endl;
+DoSpecularElectron(const G4Track& aTrack, const G4Step& aStep) {
+  G4ThreeVector k = GetLocalWaveVector(aTrack);
+  G4ThreeVector surfNorm = G4CMP::GetSurfaceNormal(aStep);
```
kelseymh, line 214: Does your branch have the "surfaceNorm" data member in G4CMPBoundaryUtils?  If so, then you can just use that here instead of recomputing it.  If you don't have that yet, we can make the change after we merge develop onto G4CMP-515.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -239,45 +246,36 @@ DoSpecularElectron(const G4ThreeVector& inDir,
     G4ThreeVector vnew = theLattice->MapPtoV_el(GetCurrentValley(),
```
kelseymh: If you are planning to cross-check the wavevector, as the `G4cout` below claims, then you should be doing `MapPtoK()` here.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -239,45 +246,36 @@ DoSpecularElectron(const G4ThreeVector& inDir,
...
-		 G4ParticleChange& /*aParticleChange*/) {
-  if (verboseLevel>1)
-    G4cout << GetProcessName() << ": Hole reflected" << G4endl;
+G4ThreeVector G4CMPDriftBoundaryProcess::
+DoSpecularHole(const G4Track& aTrack, const G4Step& aStep) {
+  if (verboseLevel>1) {
+    G4cout << GetProcessName() << "DoSpecularHole " << G4endl;
+  }
 
   G4ThreeVector surfNorm = G4CMP::GetSurfaceNormal(aStep);
```
kelseymh, line 262: See my earlier note.  If your G4CMP-513 branch include the change to {{G4CMPBoundaryUtils}} to pre-compute surfaceNorm and put it in a data member, then you can just use that data member, instead of computing it here.  If not, we can make this change later on, when merging develop into G4CMP-515.

kelseymh (changes requested):

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: No, sir. It does not.

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: k is already in the local frame, so I think I should just delete line 238.

kelseymh: Okay.  Make sure that's the case in the code that generates it (is surfNorm global or local?  Or does LambertianReflection() put it into the local frame for return?).

wlamberson1: Thank you. It looks like GetSurfaceNormal gets surfNorm in Global coordinates. For LambertianReflection, the only place I could think it would matter is when it called GetCurrentTrack()->GetPosition(). For that, I do not see any hint if it is global or local in G4CMPChargeCloud.

kelseymh: ??? Where does `G4CMPChargeCloud` come into this?  That's used optionally when creating brand new e/h tracks from an energy deposit.  I'm going to ignore that line for the rest of this.

G4Track is a Geant4 object.  Everything in it is always in global coordinates.  Therefore, track->GetPosition() will always return the position in global coordinates.

Since our `G4CMP::GetSurfaceNormal()` function _returns_ surfNorm in global coordinates, and since `LambertianReflection()` (or `GetLambertVector()`) returns a direction _relative to_ the input surface normal, those functions must be returning a global vector in this specific case.

wlamberson1: Okay, I believe I fixed it. ChargeCloud came into the picture when I looked at this LambertianReflection():

G4ThreeVector
G4CMP::LambertianReflection(const G4LatticePhysical* theLattice,
			   const G4ThreeVector& surfNorm, G4int index) {
  const G4ThreeVector surfPoint = GetCurrentTrack()->GetPosition();
  return LambertianReflection(theLattice, surfNorm, index, surfPoint);
}

I went to the declaration and definition of GetPosition() and it was in G4CMPChargeCloud.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -167,117 +168,114 @@ DoReflection(const G4Track& aTrack, const G4Step& aStep,
...
-  G4double dirNorm = inDir * surfNorm;
-  G4ThreeVector outDir = inDir - 2.*dirNorm*surfNorm;
+  // Specular reflection reverses wavevector along normal
+  G4double dirNorm = k * surfNorm;
+  k -= 2.*dirNorm*surfNorm;
+
+  // If reflected velocity is outward facing, fall back to diffuse reflection
+  G4int mode = GetPolarization(aStep.GetTrack());
+  if (!G4CMP::VelocityIsInward(theLattice, mode, k, surfNorm)) {
+    k = LambertianReflection(theLattice, surfNorm, GetCurrentValley());
```
kelseymh: If you're going to use the output of LambertianReflection as a _wavevector_, not as a transport momentum direction, then you need to test whether the momentum is inward or not, and then loop for another Lambertian attempt.  You can do that very compactly by simply changing the "if" to "while".

wlamberson1: Ah, that is a leftover from when I was incorrectly returning k instead of p. I have pushed the fix.

wlamberson1: Also, LambertianReflection() has the check-for-inward loop built in.

kelseymh (approved):
This looks good!  Once we have a validation test completed, this can be merged onto G4CMP-515.  The validation should check
* Setting the diffuse vs. specular fractions works as intended
* Individual diffuse vs. specular reflection at surfaces behave as intended

PR #72, opened 2026-02-09
G4 cmp 580

emichaud4:
Fix charge reflection so that full momentum vector is passed into FillParticleChange, rather than just a unit vector.

### DIFF
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -6,7 +6,7 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
-2026-02-09  G4CMP-580 : "Include magnitude with reflected charge momentum.
+2026-02-09  G4CMP-580 : Include magnitude with reflected charge momentum.
 2026-02-02  G4CMP-572 : Fix errors in computing lattice orientation matrix.
 2026-01-28  G4CMP-558 : Address compiler warnings from array arguments.
 
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -6,6 +6,7 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
+2026-02-09  G4CMP-580 : Reflected momentum vector needs dimensionful units.
 2026-02-02  G4CMP-572 : Fix errors in computing lattice orientation matrix.
 2026-01-28  G4CMP-558 : Address compiler warnings from array arguments.
 
--- a/ChangeHistory
+++ b/ChangeHistory
@@ -6,7 +6,7 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
-2026-02-09  G4CMP-580 : Reflected momentum vector needs dimensionful units.
+2026-02-09  G4CMP-580 : "Include magnitude with reflected charge momentum.
 2026-02-02  G4CMP-572 : Fix errors in computing lattice orientation matrix.
 2026-01-28  G4CMP-558 : Address compiler warnings from array arguments.
 
--- a/library/include/G4CMPDriftBoundaryProcess.hh
+++ b/library/include/G4CMPDriftBoundaryProcess.hh
@@ -70,7 +70,7 @@ protected:
 					   const G4ThreeVector& surfPos) const;
 
   virtual G4ThreeVector DoDiffuseElectron(const G4ThreeVector& surfNorm,
-					  const G4ThreeVector& surfPos) const;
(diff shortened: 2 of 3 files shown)


library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -199,7 +201,7 @@ DoReflectionElectron(const G4Track& aTrack, const G4Step& aStep,
     G4ThreeVector vel = GetGlobalVelocityVector(aTrack);
     reflDir = DoSpecularElectron(vel, surfNorm, surfacePoint);
   } else {
-    reflDir = DoDiffuseElectron(surfNorm, surfacePoint);
+    reflDir = DoDiffuseElectron(surfNorm, surfacePoint, ptrk);
   }
 
   FillParticleChange(GetCurrentValley(), aTrack.GetKineticEnergy(), reflDir);
```
kelseymh: I think the multiplication of `ptrk.mag()*reflDir` should be done _here_, rather than just inside `DoDiffuseElectron()`.  Notice that `DoSpecularElectron()` is also written (see below) to return a unit vector, so we want the behaviour to be consistent.

library/include/G4CMPDriftBoundaryProcess.hh
```diff
@@ -70,7 +70,7 @@ protected:
 					   const G4ThreeVector& surfPos) const;
 
   virtual G4ThreeVector DoDiffuseElectron(const G4ThreeVector& surfNorm,
-					  const G4ThreeVector& surfPos) const;
+					  const G4ThreeVector& surfPos, const G4ThreeVector& ptrk) const;
```
kelseymh: I'd suggest restoring the original interface here.  See my comment to line 207 of the .cc file.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -199,7 +201,7 @@ DoReflectionElectron(const G4Track& aTrack, const G4Step& aStep,
     G4ThreeVector vel = GetGlobalVelocityVector(aTrack);
     reflDir = DoSpecularElectron(vel, surfNorm, surfacePoint);
   } else {
-    reflDir = DoDiffuseElectron(surfNorm, surfacePoint);
+    reflDir = DoDiffuseElectron(surfNorm, surfacePoint, ptrk);
```
kelseymh: As discussed at line 207, I'd suggest restoring the original behaviour here.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -242,11 +244,12 @@ DoSpecularElectron(const G4ThreeVector& inDir,
 
 G4ThreeVector G4CMPDriftBoundaryProcess::
 DoDiffuseElectron(const G4ThreeVector& surfNorm,
-		  const G4ThreeVector& /*surfPos*/) const {
+		  const G4ThreeVector& /*surfPos*/, const G4ThreeVector& ptrk) const {
   if (verboseLevel>2) G4cout << " DoDiffuseElectron " << surfNorm << G4endl;
 
   // Charge scatters randomly off of surface
   G4ThreeVector p = G4CMP::LambertReflection(surfNorm);
+  p *= ptrk.mag();
```
kelseymh: Let's move this back to before line 207, as discussed there.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -242,11 +244,12 @@ DoSpecularElectron(const G4ThreeVector& inDir,
 
 G4ThreeVector G4CMPDriftBoundaryProcess::
 DoDiffuseElectron(const G4ThreeVector& surfNorm,
-		  const G4ThreeVector& /*surfPos*/) const {
+		  const G4ThreeVector& /*surfPos*/, const G4ThreeVector& ptrk) const {
```
kelseymh: Restore the original interface, per my discussion at line 207.

ChangeHistory
```diff
@@ -6,6 +6,7 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
+2026-02-09  G4CMP-580 : Reflected momentum vector needs dimensionful units.
```
kelseymh: This comment could be more precise.  Maybe something like "Include magnitude with reflected charge momentum."

kelseymh (changes requested):

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -20,6 +20,7 @@
 // 20251015  Resolve shadowed declaration in DoFinalReflection()
 // 20251024  G4CMP-519: Protect against possible zero energy in DoAbsorption()
 // 20251028  G4CMP-527: Move CheckStepBoundary() to ApplyBoundaryAction()
+// 20260209  G4CMP-580: Multiply refl in DoElectronDiffuse() by ptrk.mag()
```
kelseymh: Change comment to refer to DoElectronReflection().

ChangeHistory
```diff
@@ -6,6 +6,7 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
+2026-02-09  G4CMP-580 : "Include magnitude with reflected charge momentum.
```
kelseymh: Remove the extra quotation mark :-)

kelseymh (approved):

PR #73, opened 2026-03-25
Replaced Fatal Exception blocks triggered by diffusion edge cases wit…

relineha:
…h JustWarning blocks, and a killing of the track via returning negative 2D safety from the GeometryUtils functions. I'll note that these errors only occurred rarely, beyond the level of QP diffusion tested found in the validation example. It's possible more FatalException errors are lurking, but now I can go out to 2M pairbreaking energy phonons launched in a complicated qubit geometry and have it run without a fatal exception, which is a good start.

### DIFF
--- a/library/src/G4CMPGeometryUtils.cc
+++ b/library/src/G4CMPGeometryUtils.cc
@@ -377,9 +377,11 @@ G4CMP::Compute2DSafetyToDaughterVolume(const G4ThreeVector& pos,
   if (volDaughterSolid->Inside(samplePoint) == kInside) {
     G4ExceptionDescription msg;
     msg << "G4CMP::Compute2DSafetyToDaughterVolume seems to think we're "
-        << "already inside the daughter volume." << G4endl;
+        << "already inside the daughter volume. Returning negative safety "
+        << " to kill track." << G4endl;
     G4Exception("G4CMP::Compute2DSafetyToDaughterVolume()", "Geometry00X",
-                FatalException, msg);
+                JustWarning, msg);
+    return -1.0*CLHEP::um;
   }
   
   //Compute the safety. Since we should be *outside* the daughter volume
@@ -552,9 +554,12 @@ Compute2DSafetyInMotherVolume(G4VSolid * motherSolid,
     G4ExceptionDescription msg;
     msg << "G4CMP::Compute2DSafetyInMotherVolume has a returnDir whose "
         << "magnitude is zero. Given that this is swept, this should not be "
-        << "the case." << G4endl;
+        << "the case. Throwing negative safety so track gets killed" << G4endl;
     G4Exception("G4CMP::Compute2DSafetyInMotherVolume()", "Geometry00X",
-                FatalException, msg);    
+                JustWarning, msg);
+
+    //Set the safety negative, so we can pick it up with our kill-the-track flag
+    motherSafety = -1.0*CLHEP::um;
   }
   
   return motherSafety;
--- a/library/src/G4CMPGeometryUtils.cc
+++ b/library/src/G4CMPGeometryUtils.cc
@@ -167,7 +167,7 @@ G4CMP::Get2DSafetyWithDirection(const G4VTouchable* motherTouch,
     G4ExceptionDescription msg;
     msg << "G4CMP::Get2DSafetyWithDirection is returning a null direction. "
         << "Something is not running correctly.";
-    G4Exception("G4CMP::Get2DSafetyWithDirection()", "Geometry00X",
+    G4Exception("G4CMP::Get2DSafetyWithDirection()", "Geometry001",
                 FatalException, msg);
   }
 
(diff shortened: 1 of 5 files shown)


library/src/G4CMPGeometryUtils.cc
```diff
@@ -377,9 +377,11 @@ G4CMP::Compute2DSafetyToDaughterVolume(const G4ThreeVector& pos,
   if (volDaughterSolid->Inside(samplePoint) == kInside) {
     G4ExceptionDescription msg;
     msg << "G4CMP::Compute2DSafetyToDaughterVolume seems to think we're "
-        << "already inside the daughter volume." << G4endl;
+        << "already inside the daughter volume. Returning negative safety "
+        << " to kill track." << G4endl;
     G4Exception("G4CMP::Compute2DSafetyToDaughterVolume()", "Geometry00X",
```
kelseymh: Please replace the "00X" with an actual number.  Here and everywhere else.  Ideally, each G4Exception gets a different number.  If you want to replace "Geometry" with "QPGeometry" for your own clarity, that's fine too.  The string is arbitrary, but is meant to be _short_ ;-)

library/src/G4CMPGeometryUtils.cc
```diff
@@ -552,9 +554,12 @@ Compute2DSafetyInMotherVolume(G4VSolid * motherSolid,
     G4ExceptionDescription msg;
     msg << "G4CMP::Compute2DSafetyInMotherVolume has a returnDir whose "
         << "magnitude is zero. Given that this is swept, this should not be "
-        << "the case." << G4endl;
+        << "the case. Throwing negative safety so track gets killed" << G4endl;
     G4Exception("G4CMP::Compute2DSafetyInMotherVolume()", "Geometry00X",
```
kelseymh: As above, replace "00X" with a different number.

library/src/G4CMPGeometryUtils.cc
```diff
@@ -377,9 +377,11 @@ G4CMP::Compute2DSafetyToDaughterVolume(const G4ThreeVector& pos,
   if (volDaughterSolid->Inside(samplePoint) == kInside) {
     G4ExceptionDescription msg;
     msg << "G4CMP::Compute2DSafetyToDaughterVolume seems to think we're "
-        << "already inside the daughter volume." << G4endl;
+        << "already inside the daughter volume. Returning negative safety "
+        << " to kill track." << G4endl;
     G4Exception("G4CMP::Compute2DSafetyToDaughterVolume()", "Geometry00X",
-                FatalException, msg);
+                JustWarning, msg);
+    return -1.0*CLHEP::um;
```
kelseymh: Just curious: is there a reason you're using the "CLHEP::" prefix all through here?  That's not needed if we have `G4SystemOfUnits.hh` in the `#include`s.  Have you gotten "shadowing" warnings when using that .hh file?

library/src/G4CMPQPDiffusion.cc
```diff
@@ -1046,6 +1063,13 @@ G4CMPQPDiffusion::PostStepDoIt(const G4Track& track, const G4Step&) {
                            nudgedPosition,
                            track.GetMomentumDirection(),
                            false);
+
+      //Cross-check for deliberately negative safeties -- need to kill these tracks)
+      if (testSafety < 0.0) {
+        fPreemptivelyKillTrack = true;
+        fParticleChange.ProposeTrackStatus(fStopAndKill);
+        return &fParticleChange;
```
kelseymh: You've got this same code in multiple places.  That makes it a great candidate for a wrapper function.  In BoundaryUtils, I called it "DoSimpleKill()".  You can write yours with a `G4VParticleChange*` return type, and then these blocks of code become one-liners:
```
    if (testSafety < 0.) return DoSimpleKill();
```

kelseymh (changes requested):

library/src/G4CMPGeometryUtils.cc
relineha: Done. Good point -- thanks for flagging this.

library/src/G4CMPGeometryUtils.cc
relineha: Done.

library/src/G4CMPGeometryUtils.cc
relineha: Oh. Uhhh, I think this is personal habit? I can remove this.

relineha: No shadowing that I can see after removing the CLHEPs.

relineha: Though I do have to leave in `CLHEP::pi` for some reason? Guess it doesn't show up in the `G4SystemOfUnits` header.

library/src/G4CMPQPDiffusion.cc
relineha: Fair point. Done. I've also replaced a handful of other repeated-code locations where I return a dummy vector as I prep a QP kill.

PR #74, opened 2026-04-22
G4CMP-597

wlamberson1:
In order to change to UniField, the EPotFile must be set to an empty string. However, when this is attempted the following error occurs.

Illegal parameter (0) </g4cmp/EPotFile> *****

Batch is interrupted!! *****

-------- WWWW ------- G4Exception-START -------- WWWW -------

G4Exception : UIMAN0123
      issued by : G4UImanager::ApplyCommand
Command aborted (400)
Error code : 400

This is just a warning message. ***

WWWW -------- G4Exception-END --------- WWWW -------

To fix this, in the ChargeConfigMessenger.cc, make the fileCmd argument optional, and specify a default value of "" for the argument.

### DIFF
--- a/examples/charge/src/ChargeConfigMessenger.cc
+++ b/examples/charge/src/ChargeConfigMessenger.cc
@@ -10,6 +10,7 @@
 //		ChargeConfigManager.
 //
 // 20170816  Michael Kelsey
+// 20260422  G4CMP-597 -- Make fileCmd optional and provide default value.
 
 #include "ChargeConfigMessenger.hh"
 #include "ChargeConfigManager.hh"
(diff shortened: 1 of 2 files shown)


examples/charge/src/ChargeConfigMessenger.cc
```diff
@@ -34,6 +35,9 @@ ChargeConfigMessenger::ChargeConfigMessenger(ChargeConfigManager* mgr)
 
   fileCmd = CreateCommand<G4UIcmdWithAString>("EPotFile",
 			      "Set filename for non-uniform electric field");
+  // Make the filename argument optional with empty string as default
+  fileCmd->SetParameterName("filename", true);
+  fileCmd->SetDefaultValue("");
 
   hitsCmd = CreateCommand<G4UIcmdWithAString>("HitsFile",
```
kelseymh, line 42: What about here?  Should `hitsFile` also be ommittable with a null string?  Is the code where the filename gets used set up to check for an empty string and not produce output?  Or is the output file a required setting?

kelseymh (approved):
If the `hitsFile` UI command doesn't deserve modification, this is ready for merging.

examples/charge/src/ChargeConfigMessenger.cc
wlamberson1: It seems like it is a required setting. I see nothing to turn off output based on the outputFileName value.

PR #75, opened 2026-05-08
G4CMP-593: QP Diffusion Timestepper Parameterization for individual volumes

relineha:
Adding functionality to allow parameterizing the QP diffusion timestepper process for individual physical lattices.

### DIFF
--- a/library/include/G4LatticePhysical.hh
+++ b/library/include/G4LatticePhysical.hh
@@ -63,11 +63,12 @@ public:
   // Also includes setting of SC parameters. No longer ambiguous with
   // the previous one
   G4LatticePhysical(const G4LatticeLogical* Lat,
-		    G4double polycrystalElasticScatteringMFP,
-		    G4double scDelta0, G4double scTeff,
-		    G4double scDn,
-		    G4double scQPLocalTrappingTau,		    
-		    G4int h=0, G4int k=0, G4int l=0, G4double rot=0.);
+                    G4double polycrystalElasticScatteringMFP,
+                    G4double scDelta0, G4double scTeff,
+                    G4double scDn,
+                    G4double scQPLocalTrappingTau,
+                    G4double scQPDiffusionStepTau=DBL_MAX,
+                    G4int h=0, G4int k=0, G4int l=0, G4double rot=0.);
 
   
   void SetVerboseLevel(G4int vb) const {
@@ -93,6 +94,7 @@ public:
   void SetSCTeff(G4double val) { fSCTeff = val; }
   void SetSCDn(G4double val) { fSCDn = val; }
   void SetSCQPLocalTrappingTau(G4double val) { fSCQPLocalTrappingTau = val; }
+  void SetSCQPDiffusionStepTau(G4double val) { fSCQPDiffusionStepTau = val; }
 
   // Rotate input vector between lattice and solid orientations
   // Returns new vector value for convenience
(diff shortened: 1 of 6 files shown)


library/include/G4LatticePhysical.hh
```diff
@@ -63,11 +63,12 @@ public:
...
-		    G4double polycrystalElasticScatteringMFP,
-		    G4double scDelta0, G4double scTeff,
-		    G4double scDn,
-		    G4double scQPLocalTrappingTau,		    
-		    G4int h=0, G4int k=0, G4int l=0, G4double rot=0.);
+                    G4double polycrystalElasticScatteringMFP,
+                    G4double scDelta0, G4double scTeff,
+                    G4double scDn,
+                    G4double scQPLocalTrappingTau,
+                    G4double scQPDiffusionStepTau=DBL_MAX,
```
kelseymh, line 70: Do you think there are any user applications which have made use of this constructor (rather than going through G4LatticeReader)?  If so, then sticking the new parameter in the middle will break any constructors where the entries in line 71 were used.

If you don't think that's a likely issue, then this is fine.

library/src/G4CMPQPDiffusion.cc
```diff
@@ -792,7 +792,9 @@ G4VParticleChange* G4CMPQPDiffusion::AlongStepDoIt(const G4Track& track,
               << " All tries exhausted, which suggests this is perhaps a more "
               << "fundamental issue with G4CMP, or maybe your geometry is just "
               << "not following the recommended rules for tracked film "
-              << "response. Killing Track.";
+              << "response. Killing Track. fOldPosition: "
+              << fOldPosition << ", fNewPosition: " << fNewPosition
+              << " after nsteps: " << track.GetCurrentStepNumber();
```
kelseymh: On these, and generally on all of your "long message" exceptions, could you please insert line breaks (`std::endl`, not just `\n`) so that the output lines are less than about 70 characters?

kelseymh (approved):

library/include/G4LatticePhysical.hh
relineha: That scenario is one in which a user is both explicitly passing the SC parameters *and* the h, k, and l parameters, right? I don't *think* we've got enough uptake to make that likely, but it's possible that one or two people who have been using v10 may have used the SC film stuff, but I find it unlikely that anyone's actually trying to define h, k, and l in the constructor.

Also, I don't think any example code in G4CMP currently uses this constructor with explicit calls to the subsequent parameters on line 71, though, so I think it's probably okay to put this here. (Would also prefer to keep together with the other sc film parameters for organization.)

library/src/G4CMPQPDiffusion.cc
relineha: Ah, sure. I'll fix this now but then can do a more thorough sweep in a different ticket.

library/include/G4LatticePhysical.hh
kelseymh: That's fine with me.  If you take a look at what Wade has done with G4CMPSurfaceProperty (on the G4CMP-515 branch), you can see what we might have to do in your case.  For now, I think it's fine as is.

library/src/G4CMPQPDiffusion.cc
kelseymh: Works for me.  Thank you!

kelseymh (approved):

relineha:
@kelseymh I've finished cleaning up the long error comments with linebreaks. Let me know what you think here.

kelseymh (approved):

PR #79, opened 2026-05-19
G4CMP-560

jwi3:


kelseymh (changes requested):
It looks like your code change had already been included in G4CMP-585 (20260212).  I wonder if implementing a no-op destructor, so the `G4TouchableHandle` destructor is never called, might resolve the end-of-job segfault.

kelseymh:
I've converted this PR to WIP ("Draft"), until we get an answer from Geant4 about the TouchableHandle data member issue.  It might be a bug on their side, in which case, we don't need to change anything.

PR #81, opened 2026-05-20
G4 cmp 495

emichaud4:


library/include/G4CMPProcessUtils.hh
```diff
@@ -46,6 +46,7 @@
 // 20250523  Add MakePhononThetaIV for IV phonon emission angle.
 // 20250814  Add UpdatePhononWavevector() to update phonon wavevector and Vg.
 // 20251116  For G4 11, use #include "G4VTouchable.hh"
+// 20260520  Add speed of sound as argument for MakePhononEnergy.
```
kelseymh, line 49: This is in G4CMPProcessUtils, which has direct access to the lattice object.  Why does vsound have to be passed into these functions at all?  Why can't they use `theLattice->GetAverageSoundSpeed()` internally?

library/include/G4LatticeLogical.hh
```diff
@@ -200,6 +201,7 @@ public:
   void SetHoleScatter(G4double l0) { fL0_h = l0; }
   void SetHoleMass(G4double hmass) { fHoleMass = hmass; }
   void SetElectronScatter(G4double l0) { fL0_e = l0; }
+  void SetElectronLukeRateScale(G4double v) { fLukeRateScale_e = v; }
```
kelseymh: Is this parameter documented somewhere?  What does it do?  We will need to know about this "knob" for the drift-speed tuning work that our group at A&M has been doing.

library/include/G4LatticeLogical.hh
```diff
@@ -390,6 +398,7 @@ private:
   G4double fVSoundAverage;	// Speed of sound (average over phonon DOS)
   G4double fL0_e;	// Scattering length for electrons
   G4double fL0_h;	// Scattering length for holes
+  G4double fLukeRateScale_e;  // Luke scattering rate scale for electrons
```
kelseymh: Please provide a better description for this than just restating the name of the the variable.

library/src/G4CMPInterValleyRate.cc
```diff
@@ -81,27 +81,37 @@ G4double G4CMPInterValleyRate::Rate(const G4Track& aTrack) const {
...
+    }
 
     ivrate = scale * Efunc;
     IVprob.push_back(ivrate);
     totalIVRate += ivrate;
   }
 
   if (verboseLevel>2) G4cout << "IV scattering rate :  " 
-      << totalIVRate/hertz << " Hz" << G4endl;
+    << totalIVRate/hertz << " Hz" << G4endl;
```
kelseymh: Could you do this line breaking differently?  Use braces, and put the `G4cout` on the line below `if`, and use two lines for the output.

library/src/G4CMPLukeEmissionRate.cc
```diff
@@ -34,7 +35,7 @@ G4double G4CMPLukeEmissionRate::Rate(const G4Track& aTrack) const {
     return 0.;
   }
 
-  G4double kSound = 0.; G4double mass = 0.; G4double l0 = 0.; G4double vsound = 0.;
+  G4double kSound = 0.; G4double mass = 0.; G4double l0 = 0.; G4double vsound = 0.; G4double kmag = 0.;
```
kelseymh: Please break this into two lines.

library/src/G4CMPLukeEmissionRate.cc
```diff
@@ -46,6 +47,18 @@ G4double G4CMPLukeEmissionRate::Rate(const G4Track& aTrack) const {
     ktrk = theLattice->EllipsoidalToSphericalTranformation(iValley, ktrk);
     mass = sqrt(theLattice->GetElectronMass()*theLattice->GetElectronDOSMass());
```
kelseymh, line 50: Where does this expression come from?  Do we need a fourth different "GetElectronXYZMass()" function in G4LatticeLogical?

library/src/G4CMPLukeScattering.cc
```diff
@@ -205,6 +209,11 @@ G4VParticleChange* G4CMPLukeScattering::PostStepDoIt(const G4Track& aTrack,
     phi_phonon   = G4UniformRand()*twopi;
     q = 2*(kmag*cos(theta_phonon)-kSound);
 
+    if (IsElectron()) {
+      q*=1/(1-2*theLattice->GetAlpha()*theLattice->GetElectronDOSMass()*
+        vsound*vsound);
```
kelseymh: Could we make this expression easier to read, with two lines?
```
    G4double alpham = theLattice->GetAlpha()*theLattice->GetElectronDOSMass();
    q /= 1-2*alpham*vsound*vsound;
```

library/src/G4CMPProcessUtils.cc
```diff
@@ -554,12 +555,12 @@ G4double G4CMPProcessUtils::MakePhononTheta(G4double k, G4double ks) const {
 // Compute energy of phonon in Luke Scattering
 
 G4double G4CMPProcessUtils::MakePhononEnergy(G4double k, G4double ks,
-					     G4double th_phonon) const {
-  return MakePhononEnergy(2.*(k*cos(th_phonon)-ks));
+					     G4double th_phonon, G4double vsound) const {
```
kelseymh, line 558: See my comment all the way up in the .hh file.  Why do we need to pass vsound in here?  Why not have this function (or really, the next function) access theLattice->GetAverageSoundSpeed() internally?

library/src/G4LatticeLogical.cc
```diff
@@ -500,7 +513,16 @@ G4LatticeLogical::MapV_elToP(G4int ivalley, const G4ThreeVector& v_e) const {
...
-  return gamma*GetElectronMass()*c_light*v_e;
+
+  // parabolic
+  if (fAlpha*eV == 0) {
+    return gamma*GetElectronMass()*c_light*v_e;
+  }
+
+  // non-parabolic
+  else {
+    return GetElectronMass()*c_light*v_e;
```
kelseymh: Why is there no gamma factor here?

library/src/G4LatticeLogical.cc
```diff
@@ -519,7 +541,16 @@ G4LatticeLogical::MapPToP_Q(G4int ivalley, const G4ThreeVector& P) const {
...
+  // parabolic
+  if (fAlpha*eV == 0) {
+    return nToV*(GetMassTensor()*(vToN*P/GetElectronMass()));
+  }
+
+  // non-parabolic
+  else {
+    G4double nonParE = GetNonParabolicity(MapPtoEkin(ivalley,P));
+    return nToV*(GetMassTensor()*(vToN*P*nonParE/GetElectronMass()));
+  }
```
kelseymh: This could be done much more simply, without big if blocks and duplicated code:
```
   G4double nonParE = (fAlpha != 0) ? GetNonParabolicity(MapPtoEkin(ivalley,P)) : 1.;
  return nonParE*nToV*(GetMassTensor()*(vToN*P))/GetElectronMass();
```
(note that I have factored the scalars out of the nested matrix multiplications).

library/src/G4LatticeLogical.cc
```diff
@@ -616,21 +665,40 @@ G4LatticeLogical::MapEkintoP(G4int iv, const G4ThreeVector& pdir, const G4double
...
 #ifdef G4CMP_DEBUG
   if (verboseLevel>1) {
     G4cout << " <pdir|M|pdir> " << bandP << G4endl << " PMag " << PMag << G4endl 
     << " returning P " << pdir*PMag << G4endl;
   }
 #endif
 
-  return pdir*PMag;
+  // parabolic
+  if (fAlpha*eV == 0) {
```
kelseymh: See my comment above.  Please avoid big blocks of duplicated code.  You can guarantee that either you or someone else is going to make a change in one side and cause divergence.  Move everything you possibly can _out_ of the if block.  Then you can use an if statement to _just_ compute the special factor you need.

library/src/G4LatticeLogical.cc
```diff
@@ -659,8 +727,17 @@ G4LatticeLogical::MapPtoEkin(G4int iv, const G4ThreeVector& p) const {
...
+    G4double Kin = sqrt(bandP/GetElectronMass() + GetElectronMass()*c_squared*GetElectronMass()*c_squared) - GetElectronMass()*c_squared;
+    return Kin;
+  }
 
+  // non-parabolic
+  else {
+    G4double nonParE = bandP*2*GetAlpha()/GetElectronMass()/GetElectronMass()/c_squared;
+    return (sqrt(1/(1-nonParE)) -1)/2/GetAlpha();
+  }
```
kelseymh: Let's make all this conditional stuff simpler.  Move G4double Kin outside the if block.  Define local variables outside the if block
```
G4double emc2 = GetElectronMass()*c_squared;
G4double esq = emc2*emc2;
```
Now you can write:
```
if (fAlpha > 0) {
  G4double nonParE = bandP*2*fAlpha/esq;
  Kin = (sqrt(1/(1-nonParE)) - 1.) / (2*fAlpha);
} else {
  Kin = sqrt(bandP/emc2 + esq) - emc2; 
}

return Kin;
```

kelseymh (changes requested):

library/include/G4CMPProcessUtils.hh
emichaud4: Because we need AverageSoundSpeed for electron but longitudinal sound speed for holes for symmetry reasons. 

One way to see this is that for holes, the valance band is treated like a sphere which means the Deformation potential MUST BE a scalar (it's a tensor but we approximate that as a scalar for electrons). 

the probability of an hole going from state k to k' is given by : 

Image

which means this part of the equation : 

Image

reduces to simply the scalar product between the phonon momentum (q) and the phonon polarization or direction (zeta). The latter is 0 for transverse phonons because the motion of neighboring atoms is perpendicular to the propagation direction. Only longitudinal phonons are emitted by holes and thus we keep thelongitudinal sound speed in the equations

library/src/G4LatticeLogical.cc
```diff
@@ -477,7 +482,15 @@ G4LatticeLogical::MapPtoV_el(G4int ivalley, const G4ThreeVector& p_e) const {
...
 
-  return p_e*c_light/(MapPtoEkin(ivalley,p_e) + GetElectronMass()*c_squared);
+  // parabolic
+  if (fAlpha*eV == 0) {
+    return p_e*c_light/(MapPtoEkin(ivalley,p_e) + GetElectronMass()*c_squared);
+  }
+
+  // non-parabolic
+  else {
+    return p_e/ (GetElectronMass()*c_light);
```
kelseymh: This appears to be strictly non-relativistlic.  I'm quite uncomfortable with that, because we have already demonstrated that mixing relativistic (on the G4 side) vs. non-relativstic within G4CMP can drive systematic biases.

library/src/G4LatticeLogical.cc
```diff
@@ -492,15 +505,18 @@ G4LatticeLogical::MapV_elToP(G4int ivalley, const G4ThreeVector& v_e) const {
   G4double bandV = (fMassTensor.xx()*tempvec().x()*tempvec().x() +
   fMassTensor.yy()*tempvec().y()*tempvec().y() +
   fMassTensor.zz()*tempvec().z()*tempvec().z());
-  G4double gamma = 1/sqrt(1-bandV/(GetElectronMass()*c_squared));
+  // Check for parabolic or non parabolic case
+  G4double gamma = (fAlpha != 0) ? 1 : 1/sqrt(1-bandV/(GetElectronMass()*c_squared));
```
kelseymh: Why are you using non-relativistic (gamma == 1) here?  If the alpha parameter doesn't affect the calculation of gamma, then it shouldn't be triggering different kinematics.

library/src/G4LatticeLogical.cc
```diff
@@ -616,21 +644,36 @@ G4LatticeLogical::MapEkintoP(G4int iv, const G4ThreeVector& pdir, const G4double
...
+  G4double PMag = 0;
+
 #ifdef G4CMP_DEBUG
   if (verboseLevel>1) {
     G4cout << " <pdir|M|pdir> " << bandP << G4endl << " PMag " << PMag << G4endl 
     << " returning P " << pdir*PMag << G4endl;
   }
 #endif
+  // Check for parabolic or non parabolic case
+  tempvec() = (fAlpha != 0) ? pdir.unit() : pdir;
```
kelseymh: I don't understand this.  Down below, you seem to have embedded tempvec() _and_ pdir in various places where you then use the opposite if conditions to either do or don't add .unit().  Why can't you do that consistently and just either set tempvec() = pdir.unit(), or use pdir.unit() at the end everywhere.

library/src/G4LatticeLogical.cc
```diff
@@ -477,7 +482,15 @@ G4LatticeLogical::MapPtoV_el(G4int ivalley, const G4ThreeVector& p_e) const {
     G4cout << "G4LatticeLogical::MapPtoV_el " << ivalley << " " << p_e << G4endl;
 #endif
 
-  return p_e*c_light/(MapPtoEkin(ivalley,p_e) + GetElectronMass()*c_squared);
+  // parabolic
+  if (fAlpha*eV == 0) {
```
kelseymh: Don't need units when you're comparing with zero.  Maybe use `if (fAlpha <= 0.)` or swap the if and else blocks, and use `if (fAlpha > 0)` for the non-parabolic case.

library/src/G4LatticeLogical.cc
```diff
@@ -537,7 +555,17 @@ G4LatticeLogical::MapP_QToP(G4int ivalley, const G4ThreeVector& P_Q) const {
...
+  G4ThreeVector tempvec2 = P_Q;
+  tempvec2.transform(GetValley(ivalley));
+  tempvec2 /= c_light;
+
+  G4double bandP = (fMassInverse.xx()*tempvec2.x()*tempvec2.x() +
+    fMassInverse.yy()*tempvec2.y()*tempvec2.y() +
+    fMassInverse.zz()*tempvec2.z()*tempvec2.z());
+
+  // Check for parabolic or non parabolic case
+  G4double nonParE = (fAlpha != 0) ? sqrt(GetNonParabolicity(bandP)) : 1.;
```
kelseymh: I realized, after reading other code, this should probably be `(fAlpha > 0)`, so that we can use "-1" as a magic "don't do anything" code, the way we do with many other parameters.

library/src/G4LatticeLogical.cc
```diff
@@ -616,21 +644,36 @@ G4LatticeLogical::MapEkintoP(G4int iv, const G4ThreeVector& pdir, const G4double
...
+  // Check for parabolic or non parabolic case
+  tempvec() = (fAlpha != 0) ? pdir.unit() : pdir;
+  tempvec().transform(GetValley(iv));
+  bandP = (fMassTensor.xx()*tempvec().x()*tempvec().x() +
+    fMassTensor.yy()*tempvec().y()*tempvec().y() +
+    fMassTensor.zz()*tempvec().z()*tempvec().z());
 
-  return pdir*PMag;
+  // parabolic
+  if (fAlpha*eV == 0) {
```
kelseymh: Don't need units when comparing with zero.  Maybe use `if (fAlpha <= 0)` here?

library/src/G4LatticeLogical.cc
```diff
@@ -688,6 +741,16 @@ G4LatticeLogical::GetElectronEffectiveMass(G4int iv,
   return (p.mag2()-Ekin*Ekin)/(2.*Ekin*c_squared);	// Relativistic
 }
 
+// Get non-parabolic term sqrt(1+2*alpha*Ekin)
+G4double G4LatticeLogical::GetNonParabolicity(const G4double Kin) const {
+#ifdef G4CMP_DEBUG
+  if (verboseLevel>1)
+    G4cout << "G4LatticeLogical::GetNonParabolicity " << 1+2*fAlpha*Kin << G4endl;
```
kelseymh: The style of debugging message here would show the _argument_ after the name, then report what's being returned.  For example,
```
if (verboseLevel > 1) {
  G4cout << "G4LatticeLogical::GetNonParabolicity " << Kin/eV << " eV"
       << " returns " << (1+2*fAlpha*Kin)/eV << " eV" << G4endl;
}
```

library/src/G4LatticeLogical.cc
```diff
@@ -985,6 +1047,19 @@ void G4LatticeLogical::ComputeAverageSoundSpeed() {
   fVSoundAverage = fLDOS*fVSound + (fSTDOS+fFTDOS)*fVTrans;
 }
 
+// Compute Luke Scattering rate scale
+
+void G4LatticeLogical::ComputeLukeScatteringRateScale_e() {
+
+  // Merging the physical parameters in front of the rate expression 
+  // for electrons in LukeScatteringRate into one constant so we 
+  // reduce the number of multiplication/division
```
kelseymh, line 1080: This is a good explanation.  Could you put a one-line comment that summarizes this into the .hh file, for both the internal data member and the corresponding Get/Set functions?

library/src/G4LatticeLogical.cc
```diff
@@ -985,6 +1047,19 @@ void G4LatticeLogical::ComputeAverageSoundSpeed() {
...
+  fLukeRateScale_e = fAcDeform_e*fAcDeform_e*GetElectronDOSMass()*
+      GetElectronDOSMass()/GetElectronMass()/12/pi/hbar_Planck/
+      hbar_Planck/fDensity/GetAverageSoundSpeed();
```
kelseymh: In general, division is much more expensive that multiplication.  I woud recommend rewriting this so that the giant denominator is in parentheses with multiplications.  It would also be helpful for readability to factor out some of the stuff into constant expressions, e.g.,
```
G4double acDef2 = fAcDeform_e*fAcDeform_e;
G4double hbar2_12pi = 12*pi*hbar_Planck*hbarPlanck;
G4double mscale = fElectronMDOS*fElectronMDOS/fElectronMass;

fLukerateScale_e = acDef2*mscale/(fDensity*GetAverageSoundSpeed());
```

Also, does GetAverageSoundSpeed() do the calculation every time, or is there a data member?

kelseymh (changes requested):

library/include/G4CMPProcessUtils.hh
kelseymh: Ah, thank you!  Could you add a one-line comment for that in the .hh file?  Maybe something as simple as "vsound should be average of all modes for electrons, longitudinal only for holes"?

library/src/G4LatticeLogical.cc
BirgitZatschler: Since fAlpha is a G4double, I discourage the use of `==`. On some systems, it may happen that the precision for double is insufficient and this comparison may result in something like 0.0000001 == 0 and returns false. This has happened to me in another G4 application on one specific cluster, so I'm quite cautious with this. Thus, I highly encourage going for Mike's second option to swap the if and else and use `if (fAlpha > 0)`. Note, it is also possible to make a reverse condition, e.g. `if ( !(fAlpha > 0) )`.

library/src/G4CMPEqEMField.cc
```diff
@@ -198,7 +199,20 @@ void G4CMPEqEMField::EvaluateRhsGivenB(const G4double y[],
...
-  force *= fCharge*vinv*c_light;
+  if (fCharge>0) { // Temporary solution until we fix (soon) G4CMP-576.
+    force = fCharge*vinv*c_light*Efield;
+  }
+
+  if (fCharge<0) {
+    force *= fCharge*vinv*c_light;
+    nonParE = theLattice->GetNonParabolicity(theLattice->MapPtoEkin(valleyIndex,mom));
+    forceCorrection = -2*theLattice->GetAlpha()*fCharge*vinv*c_light
+      *Efield*vel*vel*theLattice->GetElectronMass()/nonParE;
```
kelseymh, line 210: These  corrections don't need any sort of if statement.  If GetAlpha() == 0, then nonParE = 1. and forceCorrection = 0., right?

library/src/G4CMPLukeEmissionRate.cc
```diff
@@ -46,6 +48,17 @@ G4double G4CMPLukeEmissionRate::Rate(const G4Track& aTrack) const {
...
     vsound = theLattice->GetAverageSoundSpeed();
+    kmag = ktrk.mag();
+    G4double Etrk = theLattice->MapPtoEkin(iValley,ptrk);
+    kSound = vsound*mass/hbar_Planck*theLattice->GetNonParabolicity(Etrk);
+    G4double qmax = 2/(1-2*theLattice->GetAlpha()*theLattice->GetElectronDOSMass()*
+      vsound*vsound)*(kmag-kSound);
+    G4double rate = theLattice->GetElectronLukeRateScale()/kmag*qmax*qmax*qmax*
+      (theLattice->GetNonParabolicity(Etrk)-3/2*theLattice->GetAlpha()*
```
kelseymh: Use "1.5" instead of "3/2"!  As written, "3/2" will do integer division and will return 1, not 1.5.

library/src/G4LatticeLogical.cc
```diff
@@ -477,7 +482,12 @@ G4LatticeLogical::MapPtoV_el(G4int ivalley, const G4ThreeVector& p_e) const {
     G4cout << "G4LatticeLogical::MapPtoV_el " << ivalley << " " << p_e << G4endl;
 #endif
 
-  return p_e*c_light/(MapPtoEkin(ivalley,p_e) + GetElectronMass()*c_squared);
+  // Check for parabolic or non parabolic case
+  if (fAlpha > 0) {
+    return p_e / (GetElectronMass() * c_light);
```
kelseymh, line 513: If this has to be non-relativistic, could you include a comment explaining why, and what other functions this non-relativistic return would cause problems for, so we don't fix it in the future?

library/src/G4LatticeLogical.cc
```diff
@@ -492,14 +502,16 @@ G4LatticeLogical::MapV_elToP(G4int ivalley, const G4ThreeVector& v_e) const {
   G4double bandV = (fMassTensor.xx()*tempvec().x()*tempvec().x() +
   fMassTensor.yy()*tempvec().y()*tempvec().y() +
   fMassTensor.zz()*tempvec().z()*tempvec().z());
-  G4double gamma = 1/sqrt(1-bandV/(GetElectronMass()*c_squared));
+  // Check for parabolic or non parabolic case
+  G4double gamma = (fAlpha > 0) ? 1 : 1/sqrt(1-bandV/(GetElectronMass()*c_squared));
```
kelseymh, line 533: As above, if this has to be non-relativistic, please put in a comment explaining why, and what other functions the return value would affect if it were changed.

library/src/G4LatticeLogical.cc
```diff
@@ -513,13 +525,15 @@ G4LatticeLogical::MapPToP_Q(G4int ivalley, const G4ThreeVector& P) const {
 
   const G4RotationMatrix& vToN = GetValley(ivalley);
   const G4RotationMatrix& nToV = GetValleyInv(ivalley);
-
+  // Check for parabolic or non parabolic case
+  G4double nonParE = (fAlpha > 0) ? GetNonParabolicity(MapPtoEkin(ivalley,P)) : 1.;
```
kelseymh: Is this necessary?  I thought GetNonParabolicity() was set up to return 1. automatically in case fAlpha == 0?

library/src/G4LatticeLogical.cc
```diff
@@ -532,12 +546,23 @@ G4LatticeLogical::MapP_QToP(G4int ivalley, const G4ThreeVector& P_Q) const {
...
+  G4ThreeVector tempvec2 = P_Q;
+  tempvec2.transform(GetValley(ivalley));
+  tempvec2 /= c_light;
+
+  G4double bandP = (fMassInverse.xx()*tempvec2.x()*tempvec2.x() +
+    fMassInverse.yy()*tempvec2.y()*tempvec2.y() +
+    fMassInverse.zz()*tempvec2.z()*tempvec2.z());
+
+  // Check for parabolic or non parabolic case
+  G4double nonParE = (fAlpha > 0) ? sqrt(GetNonParabolicity(bandP)) : 1.;
```
kelseymh: Is this actually necessary?  Doesn't GetNonParabolicity() return 1. in case alpha = 0?

library/src/G4LatticeLogical.cc
```diff
@@ -616,15 +641,26 @@ G4LatticeLogical::MapEkintoP(G4int iv, const G4ThreeVector& pdir, const G4double
...
     fMassTensor.yy()*tempvec().y()*tempvec().y() +
     fMassTensor.zz()*tempvec().z()*tempvec().z());
-  G4double PMag = sqrt(GetElectronMass()*(Ekin*Ekin+2.*Ekin*GetElectronMass()*c_squared)/(bandP));
-  
+    
+  // Check for parabolic or non parabolic case
+  if (fAlpha > 0) {
+    G4double nonParE = GetNonParabolicity(Ekin);
+    PMag = sqrt((GetElectronMass()*GetElectronMass()/2/GetAlpha()*c_squared
+      *(1-1/nonParE/nonParE ) )/(bandP));
```
kelseymh, line 683: What does this PMag simplify to in the case where alpha = 0 and nonParE = 1?  Does this reduce to the non-relativistic case?  If so, add a comment that you're doing a non-relativistic calculation here, and why.

kelseymh (changes requested):

CrystalMaps/Si/config.txt
```diff
@@ -26,7 +26,7 @@ valleyDir 1 0 0
 valleyDir 0 1 0 
 valleyDir 0 0 1
 # Intervalley scattering (matrix elements) from Jacoboni & Reggiani
```
kelseymh: This comment line is in the wrong place.  It should be after "acDeform_h".

CrystalMaps/Si/config.txt
```diff
@@ -26,7 +26,7 @@ valleyDir 1 0 0
...
-alpha 0.5 /eV
+alpha 0 /eV
```
kelseymh: Could you add three comment lines surrounding this?
* First, on the line above this, a description that this is the non-parabolicity parameter for electron transport
* Second, on this line, a comment that the best value should be 0.5, but we are keeping it zero for output consistency.
* Third, on the line below, a description of the two acDeform parameters.

CrystalMaps/Ge/config.txt
```diff
@@ -27,7 +27,7 @@ valleyDir -1 1 1
 valleyDir -1 -1 1
 valleyDir 1 -1 1
 # Intervalley scattering (matrix elements) from Jacoboni & Reggiani
-alpha 0.3 /eV
+alpha 0 /eV
```
kelseymh: See my detailed comment about this in the Si/config.txt file.

kelseymh (changes requested):

kelseymh (changes requested):
There are two residual comments, both in the box "Hidden Conversations".  One for G4CMPLukeEmissionsRate.cc, and one for G4CMPLukeScattering.cc

kelseymh (approved):

PR #82, opened 2026-05-29
G4 cmp 404

emichaud4:


library/include/G4LatticeLogical.hh
```diff
@@ -281,17 +298,34 @@ public:
...
+    return (i>=0 && i<GetNIVDeform()) ? fIVNValleys[i] : 0.;
+  }
+  G4double GetIVOrder(G4int i) const {
+    return (i>=0 && i<GetNIVDeform()) ? fIVOrder[i] : 0.;
+  }
+  G4String GetIVFGScattering(G4int i) const {
+    return (i>=0 && i<GetNIVDeform()) ? fIVFGScattering[i] : G4String("f");
+  }
+  G4String GetIVPhononMode(G4int i) const {
+    return (i>=0 && i<GetNIVDeform()) ? fIVPhononMode[i] : G4String("TA");
```
emichaud4, line 334: I'm choosing Transverse acoustic as the default value if IVphonnonmode does not pass the if condition. In G4InterValleyScattering, this will lead to the emission of either Fast or slow transverse phonon. The choice is made with a weigh (corresponding to their DOS) associated to each mode . I choose Transverse as this is the mode with the highest DOS. Should I add a comment next to this line or above?

CrystalMaps/Ge/config.txt
```diff
@@ -26,12 +26,18 @@ valleyDir 1 1 1
 valleyDir -1 1 1
 valleyDir -1 -1 1
 valleyDir 1 -1 1
-# Intervalley scattering (matrix elements)
-alpha 0.3 /eV
+# Non-parabolicty parameter for electron transport
+alpha 0 /eV     # Expected value ~0.3, kept at 0 for output consistency
```
kelseymh, line 30: Can you provide a citation for the expected value?  People want that :-)

CrystalMaps/Ge/config.txt
```diff
@@ -26,12 +26,18 @@ valleyDir 1 1 1
...
+# Acoustic deformation potentials for electrons and holes
 acDeform_e 11 eV		# Cabrera et al., arxiv(1004.1233), 2010
 acDeform_h 3.4 eV
-ivDeform 3e8 0.2e8 eV/cm		# Jacoboni & Reggiani
-ivEnergy 27.3e-3 10.3e-3 eV
+# Intervalley scattering (matrix elements) from Jacoboni & Reggiani
```
kelseymh: Is Jacoboni in arXiv?  If so, can you specify it, or give a DOI?

CrystalMaps/Si/config.txt
kelseymh, line 1: See my two comments in Ge/config.txt; they apply here also.

library/include/G4CMPCrystalGroup.hh
```diff
@@ -30,6 +30,7 @@ public:
 public:				// For convenient access to data members
   Bravais group;
   G4ThreeVector axis[3];	// Basis unit vectors in direct orientation
+  G4ThreeVector unitCell;   // Dimensions of unit cell axes
```
kelseymh, line 33: When did you add this?  I'm surprised you had to make any modifications to G4CMPCrystalGroup for what you're doing.

library/include/G4CMPProcessUtils.hh
```diff
@@ -250,8 +252,14 @@ public:
 
   // Generate direction angle for phonon generated in Luke scattering
   G4double MakePhononTheta(G4double k, G4double ks) const;
```
kelseymh: Could you rename this to MakePhononThetaLuke(), and change the call in G4CMPLukeScattering.cc?

library/include/G4CMPProcessUtils.hh
```diff
@@ -261,7 +269,7 @@ public:
 
   // Compute time between scatters/emissions for moving charge carrier
   // Parameters are "Mach number" (ratio to sound speed) and scattering length
-  G4double ChargeCarrierTimeStep(G4double mach, G4double l0) const;
+  G4double ChargeCarrierTimeStep(G4double mach, G4double l0, G4double vsound) const;
```
kelseymh: Add a comment that vsound is passed by IV depending on which scattering mode is involved.  Could you give it a default argument (-1) and fetch the average sound speed in the .cc, so that the LukeScattering code doesn't have to change?

library/include/G4LatticeLogical.hh
```diff
@@ -126,6 +128,10 @@ public:
     static const G4ThreeVector nullVec(0.,0.,0.);
     return (i>=0 && i<3 ? fBasis[i] : nullVec);
   }
+    
+  // Assign lattice constants
+  //void SetLatConst(G4double a, G4double b, G4double c);
```
kelseymh: Ah, you moved this out of G4LatticeLogical and into G4CMPCrystalGroup!  Should the Get function call through to G4CMPLatticeGroup, rather than keeping the local data member?

library/include/G4LatticeLogical.hh
```diff
@@ -235,6 +245,9 @@ public:
   const G4RotationMatrix& GetValley(G4int iv) const;
   const G4RotationMatrix& GetValleyInv(G4int iv) const;
   const G4ThreeVector& GetValleyAxis(G4int iv) const;
+    
+  // Convert eV/cm to eV for 1st order IV deformation potentials
```
kelseymh, line 248: I don't understand this (or maybe just don't understand the comment).  What intrinsic unit of length is used for this unit conversion?  Is it the size of the unit cell?

library/include/G4LatticeLogical.hh
```diff
@@ -265,6 +278,10 @@ public:
   void SetHoleAcousticDeform(G4double v) { fAcDeform_h = v; }
   void SetIVDeform(const std::vector<G4double>& vlist) { fIVDeform = vlist; }
   void SetIVEnergy(const std::vector<G4double>& vlist) { fIVEnergy = vlist; }
+  void SetIVNValleys(const std::vector<G4double>& vlist) { fIVNValleys = vlist; }
+  void SetIVOrder(const std::vector<G4double>& vlist) { fIVOrder = vlist; }
+  void SetIVFGScattering(const std::vector<G4String>& vlist) { fIVFGScattering = vlist; }
+  void SetIVPhononMode(const std::vector<G4String>& vlist) { fIVPhononMode = vlist; }
```
kelseymh, line 283: How do we ensure that all these lists are the same length?  Is there a sanity check that will generate an error if they aren't?

library/include/G4LatticeLogical.hh
```diff
@@ -281,17 +298,34 @@ public:
   G4double GetIVLinExponent() const  { return fIVLinExponent; }
 
   G4double GetAlpha() const	     { return fAlpha; }
+  G4double GetNonParabolicity(const G4double Kino) const ;
```
kelseymh: What does Kino mean?  Is that Ekin?

library/include/G4LatticeLogical.hh
```diff
@@ -311,7 +345,16 @@ private:
...
+
+  // Compute average speed of sound
+  void ComputeAverageSoundSpeed();
+
+  // Compute Luke Scattering rate scale
+  void ComputeLukeScatteringRateScale_e();
+
 private:
   // Create a thread-local buffer to use with MapAtoB() functions
   inline G4ThreeVector& tempvec() const {
```
kelseymh, line 370: If you need to have two of these temporary buffers, it would be fine to add a new tempvec2() function.

library/include/G4LatticeLogical.hh
```diff
@@ -325,6 +368,7 @@ private:
   G4String fName;			    // Name of lattice for messages
   G4CMPCrystalGroup fCrystal;		    // Symmetry group, axis unit vectors
   G4ThreeVector fBasis[3];		    // Basis vectors for Miller indices
+  G4ThreeVector fLatConst;		    // Lattice constant
```
kelseymh: See my comment above.  Since you moved the set function CrystalGroup, do we really want/need this copy?

library/src/G4CMPInterValleyRate.cc
```diff
@@ -32,89 +33,88 @@ void G4CMPInterValleyRate::
...
+    // 1st order IV rate
+    if (ivorder==1) {
+      G4double qmax = kmag*(1+sqrt(1-alpha*Emin_iv*Emin_iv/eTrk-Emin_iv/eTrk*
+      (theLattice->GetNonParabolicity(eTrk))));		// max phonon momentum
+      G4double qmin = kmag*(1-sqrt(1- alpha*Emin_iv*Emin_iv/eTrk-Emin_iv/eTrk*
+      (theLattice->GetNonParabolicity(eTrk))));		// min phonon momentum
+      scale = m_DOS3half*m_DOS * nVal * D_iv*D_iv
+        /(16*pi*hbar_Planck*density*theLattice->GetElectronMass()*
+        sqrt(theLattice->GetElectronMass())*Emin_iv*kmag)*
+        (theLattice->GetNonParabolicity(eTrk)-2*alpha*Emin_iv);
```
kelseymh, line 104: Is this expression correct?  The way the parentheses are written, the "16*pi....*kmag" is in the denominator, but "NonPar - 2*alpha*Emin_iv" is in the _numerator.

library/src/G4CMPInterValleyScattering.cc
```diff
@@ -151,6 +171,230 @@ G4CMPInterValleyScattering::PostStepDoIt(const G4Track& aTrack,
...
+    std::vector<G4double> probabilities = ivpro->GetIVProb();
+    std::vector<G4double> cumulatives(probabilities.size());
+    G4double totalIVRate = 0.;
+
+    // Calculate total IV rate
+    for (auto& element : probabilities) { totalIVRate += element; }
+
+    // Don't do anything if IV rate is 0 
+    if (totalIVRate == 0) { 
+      if (verboseLevel > 1) {G4cout << "NO IV Rate " << G4endl;}
```
kelseymh: You can remove the braces for this one-line statement.

library/src/G4CMPInterValleyScattering.cc
```diff
@@ -151,6 +171,230 @@ G4CMPInterValleyScattering::PostStepDoIt(const G4Track& aTrack,
...
+  }
+
+  // Get IV rates 
+  else if (ivpro != nullptr) {
+    std::vector<G4double> probabilities = ivpro->GetIVProb();
+    std::vector<G4double> cumulatives(probabilities.size());
+    G4double totalIVRate = 0.;
+
+    // Calculate total IV rate
+    for (auto& element : probabilities) { totalIVRate += element; }
```
kelseymh: I think this could be done with an STL algorithm instead of the for loop.

library/src/G4CMPInterValleyScattering.cc
```diff
@@ -151,6 +171,230 @@ G4CMPInterValleyScattering::PostStepDoIt(const G4Track& aTrack,
...
+  // Make sure it's not empty
+  if (ivpro == nullptr) { 
+    if (verboseLevel > 1) {G4cout << "NO G4CMPInterValleyRate* " << G4endl;}
+    return &aParticleChange;
+  }
+
+  // Get IV rates 
+  else if (ivpro != nullptr) {
+    std::vector<G4double> probabilities = ivpro->GetIVProb();
+    std::vector<G4double> cumulatives(probabilities.size());
```
kelseymh: I would move this line down below to where it's used.

library/src/G4CMPInterValleyScattering.cc
```diff
@@ -151,6 +171,230 @@ G4CMPInterValleyScattering::PostStepDoIt(const G4Track& aTrack,
...
+    for (auto& element : probabilities) { totalIVRate += element; }
+
+    // Don't do anything if IV rate is 0 
+    if (totalIVRate == 0) { 
+      if (verboseLevel > 1) {G4cout << "NO IV Rate " << G4endl;}
+      return &aParticleChange;
+    }
+
+    // Do IV scattering if IV rate is not 0 
+    else if (totalIVRate !=0) {
```
kelseymh: This else is unnecessary, since the "if == 0" block has a return.  Remove it and move the contents out one level.

library/src/G4CMPInterValleyScattering.cc
```diff
@@ -151,6 +171,230 @@ G4CMPInterValleyScattering::PostStepDoIt(const G4Track& aTrack,
...
+
+      // Normalize IV rates 
+      for (auto& element : probabilities) { element /= totalIVRate; }
```
kelseymh: This can be done with an STL algorithm, and I think the cumulative sum can as well.  We have a lot of this in the SuperSim CDMSsources handlers.

library/src/G4CMPInterValleyScattering.cc
```diff
@@ -151,6 +171,230 @@ G4CMPInterValleyScattering::PostStepDoIt(const G4Track& aTrack,
...
+        k_recoilHV = theLattice->EllipsoidalToSphericalTranformation(valley, k_recoil);           
+
+        // IV f or g-type scattering
+        G4String ivfgscat = theLattice->GetIVFGScattering(Ephononi); 
+
+        // Find anti-valley
+        G4int antivalley = 0;
+
+        if (valley%2==0) {antivalley=valley+1;}
+        if (valley%2!=0) {antivalley=valley-1;}
```
kelseymh: `G4int antivalley = valley + (1 - 2*(valley%2));`

One line, no if conditions needed.

library/src/G4CMPLukeEmissionRate.cc
```diff
@@ -34,9 +35,9 @@ G4double G4CMPLukeEmissionRate::Rate(const G4Track& aTrack) const {
     return 0.;
   }
 
-  G4double kSound = 0.; G4double mass = 0.; G4double l0 = 0.;
+  G4double kSound = 0.; G4double mass = 0.; G4double l0 = 0.;
```
kelseymh: There's an extra space after the last semicolon.  If you delete it, this line will show as unchanged from the develop version.

library/src/G4CMPLukeEmissionRate.cc
```diff
@@ -45,20 +46,34 @@ G4double G4CMPLukeEmissionRate::Rate(const G4Track& aTrack) const {
     // Turning wavevector to spherical frame where electrons act like holes
     // as the mass is isotropic
     ktrk = theLattice->EllipsoidalToSphericalTranformation(iValley, ktrk);
-    mass = theLattice->GetElectronMass();
-    // The l0 in configuration file is calculated using the conductivity mass
+    // Mass expression comes from our approximation q = sqrt(md/mc) q*
```
kelseymh, line 50: Insert a blank line above this comment.

library/src/G4CMPLukeScattering.cc
```diff
@@ -137,17 +138,26 @@ G4VParticleChange* G4CMPLukeScattering::PostStepDoIt(const G4Track& aTrack,
...
+  G4double kSound = 0.;
   if (IsElectron()) {
     ktrk = lat->MapPtoK(iValley, ptrk);
     // Turning wavevector to spherical frame where electrons act like holes
     // as the mass is isotropic
     ktrk = lat->EllipsoidalToSphericalTranformation(iValley, ktrk);
-    mass = lat->GetElectronMass();
+    // Mass expression comes from our approximation q = sqrt(md/mc) q*
+    mass = sqrt(theLattice->GetElectronMass()*theLattice->GetElectronDOSMass());
+    vsound = theLattice->GetAverageSoundSpeed();
```
kelseymh, line 150: Move this below Etrk to match what's done in the other block.

library/src/G4LatticeLogical.cc
```diff
@@ -221,13 +230,23 @@ void G4LatticeLogical::SetCrystal(G4CMPCrystalGroup::Bravais group, G4double a,
...
 				  G4double beta, G4double gamma) {
   fCrystal.Set(group, alpha, beta, gamma);	// Defines unit cell axes
+  fCrystal.SetUnitCell(a,b,c);    // Define unit cell lattice parameters
 
   fBasis[0] = a*fCrystal.axis[0];	// Basis vectors include spacing
   fBasis[1] = b*fCrystal.axis[1];
   fBasis[2] = c*fCrystal.axis[2];
+
+  fLatConst = fCrystal.unitCell;
+
```
kelseymh: Why not just change GetLatConst() to call through to fCrystal.GetUnitCell()?

library/src/G4LatticeLogical.cc
```diff
@@ -221,13 +230,23 @@ void G4LatticeLogical::SetCrystal(G4CMPCrystalGroup::Bravais group, G4double a,
...
 }
 
 
+// void G4LatticeLogical::SetLatConst(G4double a, G4double b, G4double c) {
+//   fCrystal.SetLatConst(a,b,c);	// Define unit cell lattice parameters
```
kelseymh: Why drop this part?  It's a convenient call through if that's needed somewhere.

library/src/G4LatticeLogical.cc
```diff
@@ -507,12 +573,22 @@ G4LatticeLogical::MapP_QToP(G4int ivalley, const G4ThreeVector& P_Q) const {
...
+  G4double nonParE = sqrt(GetNonParabolicity(bandP));
+
 #ifdef G4CMP_DEBUG
   if (verboseLevel>1) 
-    G4cout << " P " << nToV*(GetMInvTensor()*(vToN*P_Q*electron_mass_c2/c_squared)) << G4endl;
+    G4cout << " P " << nToV*(GetMInvTensor()*(vToN*P_Q/nonParE*GetElectronMass())) << G4endl;
 #endif
 
-  return nToV*(GetMInvTensor()*(vToN*P_Q*GetElectronMass()));
+  return nToV*(GetMInvTensor()*(vToN*P_Q/nonParE*GetElectronMass()));
```
kelseymh: It would be much less confusing in these lines if you wrote
```
(vToN*P_Q*GetElectronMass()/nonParE)
```
It's very easy to read that and _think_ the whole expresssion is in the denominator, but it's not.

library/src/G4LatticeLogical.cc
```diff
@@ -522,8 +598,8 @@ G4LatticeLogical::MapV_elToK(G4int ivalley, const G4ThreeVector &v_e) const {
     G4cout << "G4LatticeLogical::MapV_elToK " << ivalley << " " << v_e << G4endl;
 #endif
 
-  tempvec() = MapV_elToP(ivalley, v_e);
-  return MapPtoK(ivalley, tempvec());
+  G4ThreeVector p = MapV_elToP(ivalley, v_e);
```
kelseymh, line 600: You did this because of the nested tempvec() usage?

library/src/G4LatticeLogical.cc
```diff
@@ -770,7 +882,7 @@ void G4LatticeLogical::SetMassTensor(const G4RotationMatrix& etens) {
 void G4LatticeLogical::FillMassInfo() {
   // Effective mass for conductivity calculations
   fElectronMass = 3. / ( 1./fMassTensor.xx() + 1./fMassTensor.yy()
-			 + 1./fMassTensor.zz() );  
+			 + 1./fMassTensor.zz() );
```
kelseymh: Thank you for cleaning this up.  There's still one space after the semicolon, if you want to remove it.

library/src/G4LatticeLogical.cc
```diff
@@ -865,6 +977,20 @@ void G4LatticeLogical::AddValley(const G4RotationMatrix& valley) {
...
 }
 
+void G4LatticeLogical::ConverteVcmToeV(const std::vector<G4double>& ivrateorder) {
+
+std::vector<G4double> ivdeformtest = GetIVDeform();
+G4ThreeVector latticeconstant = GetLatConst();
+
+for (size_t i = 0; i < ivrateorder.size(); i++) {
+    if (ivrateorder[i] == 1) {
+    ivdeformtest[i]=ivdeformtest[i]*latticeconstant.x();
```
kelseymh: Write this as `ivdeformtest[i] *= latticeconstant.x();`

library/src/G4LatticeLogical.cc
```diff
@@ -865,6 +977,20 @@ void G4LatticeLogical::AddValley(const G4RotationMatrix& valley) {
   fValleyAxis.push_back(fValleyInv.back()*G4ThreeVector(1.,0.,0.));
 }
 
+void G4LatticeLogical::ConverteVcmToeV(const std::vector<G4double>& ivrateorder) {
+
+std::vector<G4double> ivdeformtest = GetIVDeform();
+G4ThreeVector latticeconstant = GetLatConst();
```
kelseymh: If you only need the x component below, why not write this as
```
G4double latConstX = GetLatConst().x();
```
and then use `*= latConstX;` below?

library/src/G4LatticeLogical.cc
```diff
@@ -865,6 +977,20 @@ void G4LatticeLogical::AddValley(const G4RotationMatrix& valley) {
...
+
+for (size_t i = 0; i < ivrateorder.size(); i++) {
+    if (ivrateorder[i] == 1) {
+    ivdeformtest[i]=ivdeformtest[i]*latticeconstant.x();
+    }    
+}
+SetIVDeform(ivdeformtest);
+
```
kelseymh, line 996: These lines look like the have the wrong indentation.

library/src/G4LatticeLogical.cc
```diff
@@ -917,23 +1043,50 @@ const G4ThreeVector& G4LatticeLogical::GetValleyAxis(G4int iv) const {
...
 
-G4double G4LatticeLogical::ComputeL0(G4bool IsElec) {
+void G4LatticeLogical::ComputeL0(G4bool IsElec) {
   G4double mass = 0.;
   G4double acDeform = 0.;
-      
+
   if (IsElec) {
-      mass = GetElectronMass();
+      mass = GetElectronDOSMass();
```
kelseymh: I wonder if _this_ is why you're seeing substantially different drift speed.  This will greatly affect l0.  Does it also affect the other calculations where you assumed md/mc etc. in the LukeScattering code?

ChangeHistory
```diff
@@ -6,8 +6,20 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
-2026-05-28  g4cmp-V10-02-00 Several bug fixes and EnergyPartition improvements
-2026-05-28  G4CMP-617 : Simplify nParticlesMinimum tests in EnergyPartition
+[ Modifications included on branch G4CMP-404, put subtask tags below it: ]
+
+2026-05-26  G4CMP-404 : Improve charge carriers-phonons scattering.
```
kelseymh: Don't forget to update this last date when we're ready to merge :-)  I'll also ask you, once all other comments have been resolved, to remove lines 9-10 and 20-21.

ChangeHistory
```diff
@@ -98,6 +110,7 @@ ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 2025-05-05  G4CMP-479 : Update phonon time to account for surface displacement.
 2025-04-30  G4CMP-461 : Add ability to skip over flats in specular reflections.
 2025-04-28  G4CMP-465 : Implement G4CMPSolidUtils for custom solid functions.
+
```
kelseymh: I think you introduced this by mistake.  Could you remove it?

kelseymh (changes requested):

CrystalMaps/Ge/config.txt
emichaud4: Done. I used the same style as the other citations. Let me know if this is okay (one line is a bit long)

CrystalMaps/Ge/config.txt
emichaud4: I used the same style as the other citations but I can also paste the DOI instead if you prefer : https://doi.org/10.1103/RevModPhys.55.645

Let me know which one you prefer!

CrystalMaps/Si/config.txt
emichaud4: Same comments as Si about the style

library/include/G4CMPCrystalGroup.hh
emichaud4: I added this a couple of months ago if not more than a year ago. I could not find the lattice parameter (~5.5 Angstrom) but needed it to convert 1st order IV deformation potential from eV to eV/cm

library/include/G4CMPProcessUtils.hh
emichaud4: Done!

library/include/G4CMPProcessUtils.hh
emichaud4: I realized that the Luke rate for electron is now calculated without this function, meaning I can revert it back to longitudinal speed of sound because ChargeCarrierTimeStep is only used by holes.

emichaud4: I included a -1 default value and I'm using longitudinal speed of sound as default because only holes use this function for which we must use longitudinal speed of sound for symmetry/group reasons

emichaud4: I added a comment as well above ChargeCarrierTimeStep

library/include/G4LatticeLogical.hh
emichaud4: I did something similar to fBasis. I assigned fLatConst = fCrystal.unitCell; in void G4LatticeLogical::SetCrystal. I removed the commented void SetLatConst.

library/include/G4LatticeLogical.hh
emichaud4: 0th order deformation potential have eV/cm units while 1st order have eV units. Because I was not able to use different units on the same line in config.txt, users input the deformation potentials in eV/cm and a conversion is done after for 1st order deformation potentials. Let me know if there is a better way to do this

emichaud4: For Si and Ge, this is done easily because the lattice constant is the same in all direction, so we can use any of the direction (x, y or z) of the lattice constant

library/include/G4LatticeLogical.hh
emichaud4: I added a function CheckIVConsistency to make sure the lists are the same size

library/include/G4LatticeLogical.hh
emichaud4: oups, I must have changed that for testing at one point and forgot to rename it back to Ekin. I have corrected the mistake

library/include/G4LatticeLogical.hh
emichaud4: Good idea, I created a tempvec2. I need it in 2 of the map functions

library/include/G4LatticeLogical.hh
emichaud4: I tried to copy the method used for fBasis just above. I'm not sure how I could remove fLatConst.

library/src/G4CMPInterValleyRate.cc
emichaud4: yes, it's supposed to be this way : 

Image

(we can forget about Nq, it is =0 at T=0K)

library/src/G4CMPInterValleyScattering.cc
emichaud4: Done!

library/src/G4CMPInterValleyScattering.cc
emichaud4: I'm not sure where to look for those STL algorithms. I have looked quickly in SuperSim CDMSSources into some .cc files but could not find anything. Can you, or maybe Birgit, point me to a file in particular? Thank you!

library/src/G4CMPInterValleyScattering.cc
emichaud4: See my other comment about SuperSim CDMSSources

library/src/G4CMPInterValleyScattering.cc
emichaud4: Done!

library/src/G4CMPInterValleyScattering.cc
emichaud4: Done!

library/src/G4CMPInterValleyScattering.cc
emichaud4: Done!

library/src/G4CMPLukeEmissionRate.cc
emichaud4: Removed

library/src/G4CMPLukeEmissionRate.cc
emichaud4: Done!

library/src/G4CMPLukeScattering.cc
emichaud4: done!

library/src/G4LatticeLogical.cc
emichaud4: I'm not sure what I was aiming for with this part. Wouldn't that be the equivalent of fCrystal.UnitCell?

library/src/G4LatticeLogical.cc
emichaud4: Found an easy way to do so finally :) ! I'll remove the fLatConst since it's not needed

library/src/G4LatticeLogical.cc
emichaud4: Yes, I confirmed that fCrystal.unitCell.x() and a are the same thing

library/include/G4LatticeLogical.hh
emichaud4: Finally, I got rid of all fLatConst and even the GetLatticeConst() function. I instead use fCrystal.unitCell which does the same thing!

library/src/G4LatticeLogical.cc
emichaud4: I'll drop that part finally as I can get what I need from fCrystal.unitCell()

library/src/G4LatticeLogical.cc
emichaud4: Done!

library/src/G4LatticeLogical.cc
emichaud4: Yes, using both tempvec2() or tempvec() crashes the simulation. I can try a workaround if you prefer not to use p!

library/src/G4LatticeLogical.cc
emichaud4: Done!

library/src/G4LatticeLogical.cc
emichaud4: I removed all flatConst and GetLatConst and instead use fCrystal.unitCell

library/src/G4LatticeLogical.cc
emichaud4: I'm now using : 
`ivdeformtest[i]=ivdeformtest[i]*fCrystal.unitCell.x();`

library/src/G4LatticeLogical.cc
emichaud4: Corrected!

library/src/G4LatticeLogical.cc
emichaud4: Yes, this is probably the reason why we see such a different drift speed (and also using the mean speed of sound). I'm not against using mc instead of md as it is done in develop and master. I felt like md is a better approximation than mc for (especially for Ge):

[massapprox.pdf](https://github.com/user-attachments/files/28754721/massapprox.pdf)

emichaud4: md is also the one used in literature and I was hoping that tuning the deformation potentials, which varies a lot, would "counter" the effect of using md instead of mc. Because all those parameters are highly correlated (speed of sound, deformation potentials and masses) when fitting drift speed to G4CMP, I don't mind reverting back to mc. If we choose to do so, I'll just have to modify a little bit G4CMPLukeScattering and Rate at other places

emichaud4: My main logic was really that ther average speed of sound and md are more "physically justified" and that we could tune Dac after

emichaud4: Tell me what you think Mike!

ChangeHistory
emichaud4: Yes, I'll do that just before the merge! I expect a couple other changes to G4CMP-404 so I'll only do the switch at the end if that's okay with you?

ChangeHistory
emichaud4: REmoved the line

library/src/G4LatticeLogical.cc
emichaud4: here is the mass/speed of sound table for comparaison : 

[masstables.pdf](https://github.com/user-attachments/files/28755136/masstables.pdf)

library/include/G4LatticeLogical.hh
```diff
@@ -276,12 +272,31 @@ public:
...
   void SetHoleAcousticDeform(G4double v) { fAcDeform_h = v; }
-  void SetIVDeform(const std::vector<G4double>& vlist) { fIVDeform = vlist; }
-  void SetIVEnergy(const std::vector<G4double>& vlist) { fIVEnergy = vlist; }
-  void SetIVNValleys(const std::vector<G4double>& vlist) { fIVNValleys = vlist; }
-  void SetIVOrder(const std::vector<G4double>& vlist) { fIVOrder = vlist; }
-  void SetIVFGScattering(const std::vector<G4String>& vlist) { fIVFGScattering = vlist; }
-  void SetIVPhononMode(const std::vector<G4String>& vlist) { fIVPhononMode = vlist; }
+  void SetIVDeform(const std::vector<G4double>& vlist) {
+    fIVDeform = vlist;
+    CheckIVConsistency();
```
kelseymh: I don't think this is a safe thing to do in each individual function.  If you think about it, the first call will almost certainly put things into an inconsistent state, because all the other parameters aren't set yet.

Ryan addressed this kind of thing by having a Set function with a list of parameters, so the user could call it with all of them.

If these functions are being called from G4LatticeReader, then I would remove all the CheckIVConsistency() calls here, and put just a single call to that in the Initialize() function.

library/include/G4LatticeLogical.hh
```diff
@@ -128,10 +128,6 @@ public:
     static const G4ThreeVector nullVec(0.,0.,0.);
     return (i>=0 && i<3 ? fBasis[i] : nullVec);
   }
-    
-  // Assign lattice constants
-  //void SetLatConst(G4double a, G4double b, G4double c);
-  const G4ThreeVector& GetLatConst() const { return fLatConst; }
```
kelseymh: Is it safe to remove this accessor?  If there's other code that uses it, especially out in user applications, I think I'd make it a call-through to the `G4CMPCrystalGroup` equivalent function.

library/include/G4LatticePhysical.hh
```diff
@@ -139,9 +139,6 @@ public:
 
   // Call through to get crystal basis vectors
   const G4ThreeVector& GetBasis(G4int i) const { return fLattice->GetBasis(i); }
-    
-  // Get Lattice constant
-  const G4ThreeVector& GetLatConst() const { return fLattice->GetLatConst(); }
```
kelseymh: If you agree to my comment in G4LatticeLogical.hh, this should be restored.

library/src/G4CMPLukeEmissionRate.cc
```diff
@@ -46,11 +46,12 @@ G4double G4CMPLukeEmissionRate::Rate(const G4Track& aTrack) const {
     // Turning wavevector to spherical frame where electrons act like holes
     // as the mass is isotropic
     ktrk = theLattice->EllipsoidalToSphericalTranformation(iValley, ktrk);
+
     // Mass expression comes from our approximation q = sqrt(md/mc) q*
     mass = sqrt(theLattice->GetElectronMass()*theLattice->GetElectronDOSMass());
-    vsound = theLattice->GetAverageSoundSpeed();
     kmag = ktrk.mag();
     G4double Etrk = theLattice->MapPtoEkin(iValley,ptrk);
+    vsound = theLattice->GetAverageSoundSpeed();
```
kelseymh, line 54: Why move this line?  Just for readability?  If so, that's fine.

library/src/G4CMPLukeEmissionRate.cc
```diff
@@ -68,12 +69,12 @@ G4double G4CMPLukeEmissionRate::Rate(const G4Track& aTrack) const {
     mass = theLattice->GetHoleMass();
     vsound = theLattice->GetSoundSpeed();
   }
-
   kmag = ktrk.mag();
+
```
kelseymh, line 74: I'd rather have the blank line after the close-brace.

library/src/G4LatticeLogical.cc
```diff
@@ -980,17 +969,35 @@ void G4LatticeLogical::AddValley(const G4RotationMatrix& valley) {
...
+  const size_t n = fIVDeform.size();
+
+  if (fIVEnergy.size() != n ||
+      fIVNValleys.size() != n ||
+      fIVOrder.size() != n ||
+      fIVFGScattering.size() != n ||
+      fIVPhononMode.size() != n)
+  {
+    G4Exception("G4LatticeLogical::CheckIVConsistency",
+                "IV_SIZE_MISMATCH",
```
kelseymh: The Geant4 convention for error codes is a shortname and a number.  Here, I would use "Lattice001" or "Lattice002" or something.

kelseymh (changes requested):

library/src/G4CMPLukeEmissionRate.cc
emichaud4: because we need a different vsound for electrons and holes. vsound used to be calculated outside the if loops because it was the same for both charges.

library/include/G4LatticeLogical.hh
emichaud4: Done! I now have only one CheckIVConsistency call in Initialize().

library/src/G4CMPLukeEmissionRate.cc
emichaud4: Moved!

library/src/G4LatticeLogical.cc
emichaud4: changed it to Lattice002 since 001 was already taken

CrystalMaps/Ge/config.txt
BirgitZatschler: If there is a DOI, that would be my preference. If there is none, an arXiv identifier is also great. Just names without a year or journal title could be ambiguous and hard to find.

library/src/G4CMPInterValleyScattering.cc
BirgitZatschler: I guess Mike is referring to using STL algoritms for the vector `probabilities`, see e.g. [https://www.geeksforgeeks.org/cpp/cpp-stl-cheat-sheet/](https://www.geeksforgeeks.org/cpp/cpp-stl-cheat-sheet/)
In principle you don't need to define `element`, but could change the entries of `probabilities` directly, _if this works in this case_.

One example for [egam](https://gitlab.com/supercdms/Simulations/supersim/-/blob/master/CDMSsources/CDMSEventGenReader.cc?ref_type=heads#L85-87:~:text=egam%2Epush%5Fback) and one for [primaries](https://gitlab.com/supercdms/Simulations/supersim/-/blob/master/CDMSsources/CDMSEventGenData.cc?ref_type=heads#L51-54). If Mike had something different in mind, please correct me.

library/src/G4LatticeLogical.cc
kelseymh: Thank you, Emile!  If you can work with @djcranshaw on the acDeform tuning to get the right drift speeds, that would be really great.  I would really prefer that G4CMP-404 end up with consistent observables when we release it.

ChangeHistory
kelseymh: Perfect!

library/include/G4LatticeLogical.hh
kelseymh: I think I'd put the way you're chosing and why on the line above each function declaration, along with the purpose of the Get function.

library/include/G4LatticeLogical.hh
emichaud4: This is not used anywhere else. It's something I added a couple months ago myself while trying to find the best way to get the lattice parameter to convert some deformation potential units.

CrystalMaps/Ge/config.txt
emichaud4: Changed the style to use DOI instead

library/src/G4CMPInterValleyScattering.cc
emichaud4: I have changed all the probabilities, cumulatives and total rates calculations so it uses an STL algorithm. Tell me if this is what you had in mind!

library/src/G4CMPInterValleyScattering.cc
emichaud4: I have changed all the probabilities, cumulatives and total rates calculations so it uses an STL algorithm. Tell me if this is what you had in mind!

library/include/G4LatticeLogical.hh
emichaud4: I added a couple of comments above the functions. Let me know what you think!

library/src/G4CMPConfigMessenger.cc
```diff
@@ -46,6 +46,7 @@
 // 20260429  G4CMP-598: Add macro command for minimum particle generation.
 // 20250502  G4CMP-358: Add macro command for maximum steps (stuck tracks).
 // 20250527  G4CMP-485: Change IVRate name to Matrix.
+// 20260606  G4CMP-578: Add macro command for pprimary phonon energy.
```
kelseymh: Fix spelling (pprimary -> primary)

library/src/G4CMPInterValleyRate.cc
```diff
@@ -13,6 +13,7 @@
 //		acoustic rate, as it is _intra_valley.
 // 20170919  Add interface for threshold identification
 // 20250423  Add parabolic ellipsoidal bands IV scattering rate.
+// 20260618  G4CMP-628 -- Skip IV scattering (MFP=0) for single valley case.
```
kelseymh: Technically, it's _rate_=0, so MFP=DBL_MAX.

library/src/G4CMPInterValleyScattering.cc
```diff
@@ -219,14 +220,12 @@ G4CMPInterValleyScattering::ValleyScattering(const G4Track& aTrack,
     }
 
     // Normalize IV rates 
-    for (auto& element : probabilities) { element /= totalIVRate; }
+    std::transform(probabilities.begin(), probabilities.end(), probabilities.begin(),
```
kelseymh, line 224: Just curious, is `std::divides` deprecated?  In SuperSim (e.g., CDMSsources/CDMSBinnedData.cc) we used `std::bind(std::divides<G4double>(), _1, sum)`.

ChangeHistory
```diff
@@ -20,10 +20,30 @@ ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
...
+2026-06-12  G4CMP-620 : Expanded quasiparticle example for use at RISQ 2026.
+2026-06-06  G4CMP-578 : New parameter for primary phonon energy partitioning.
+
+2026-06-05  g4cmp-V10-03-00  Avoid FatalExceptions, add QPDiffusion setting.
+2026-06-03  G4CMP-619 : Softened some exception types in Boundary/GeometryUtils
+2026-06-01  G4CMP-593 : Add diffusion time stepper tau for individual lattices
+
+2026-05-28  g4cmp-V10-02-00 Several bug fixes and EnergyPartition improvements
+2026-05-28  G4CMP-617 : Simplify nParticlesMinimum tests in EnergyPartition
+>>>>>>> origin/develop
```
kelseymh: Delete this line :-)

library/src/G4CMPConfigMessenger.cc
emichaud4: oups! corrected!

library/src/G4CMPInterValleyRate.cc
emichaud4: changed to rate=0

library/src/G4CMPInterValleyScattering.cc
emichaud4: changed it to your suggestion!

ChangeHistory
emichaud4: done!

ChangeHistory
```diff
@@ -38,7 +39,7 @@ ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
 2026-05-28  g4cmp-V10-02-00 Several bug fixes and EnergyPartition improvements
 2026-05-28  G4CMP-617 : Simplify nParticlesMinimum tests in EnergyPartition
->>>>>>> origin/develop
+
```
kelseymh: Remove this blank line; all these tags are part of the V10-02-00 release.

kelseymh (changes requested):
Looks great!  We repeat the tuning that Derek did for acDeform in germanium, to get us back to the good drift speed performance.  I think we should also "tune" the silicon acDeform, not to data, but rather just so G4CMP-404 produces whatever drift-speed shape we currently have with V10-04-01.

It would also be good for you, Derek and Haley to coordinate how the three IV models' parameter sets might be tuned to match the two-dimensional data from the Moffatt (Cabrera) papers.  I'd like to get a clearer sense of how much time that project could take, in order to decide wether to do it on the G4CMP-404 branch, or as a separate effort.

ChangeHistory
emichaud4: Done!

PR #83, opened 2026-05-29
G4CMP-515

wlamberson1:
The situation with charge reflection in [G4CMP-503](https://jira.slac.stanford.edu/browse/G4CMP-503) has turned out to be a larger and more complex task than originally expected. I propose that we roll out what has currently been implemented (fixed DriftBoundaryProcess::DoAbsorption() logic, diffuse-only scattering for electrons), so that experiments which need charge reflection can proceed. The remaining tasks (added here as subtasks) can be implemented in a more structured way, without keeping users with immediate needs on hold.

This larger task was driven by a set of three questions I asked in [G4CMP-503](https://jira.slac.stanford.edu/browse/G4CMP-503):

Is reflecting the velocity vector (which I think is equivalent to k, in this case?) the right thing to do?
Is specular reflection for the charges the right thing to do at all? If they reach the surface and don't get trapped, should they "reflect", or just scatter randomly?
Should we be reassigning the valley? If the charge direction is getting significantly flipped, it will be far outside its assigned valley (indeed, it will be in the opposite hemisphere!). We don't do that reassignment for NTL, but the latter is a smaller deviation (less than 90 degrees).


examples/charge/src/ChargeDetectorConstruction.cc
```diff
@@ -88,6 +91,14 @@ G4VPhysicalVolume* ChargeDetectorConstruction::Construct()
...
     }
 
+    // Check for changes in surface probabilities (new)
+    if (chargeAbsorbProb != ChargeConfigManager::GetChargeAbsorbProb() ||
+        specularReflectProb != ChargeConfigManager::GetSpecularReflectProb()) {
+      delete topSurfProp; topSurfProp = nullptr;
+      delete botSurfProp; botSurfProp = nullptr;
+      delete wallSurfProp; wallSurfProp = nullptr;
+    }
+
```
kelseymh: You don't really need to do this.  `G4CMPSurfaceProperty::FillChargeMaterialPropertiesTable(abs, refl, spec, eMink, hMinK)` is public, so you can call it for the already existing SurfaceProperty objects, and just update the data values.

library/include/G4CMPSurfaceProperty.hh
```diff
@@ -116,9 +165,9 @@ public:
   }
 
   // Functions to compute reflection probabilities vs. frequency
-  G4double AnharmonicReflProb(G4double freq) const;
-  G4double DiffuseReflProb(G4double freq) const;
-  G4double SpecularReflProb(G4double freq) const;
+  G4double AnharmonicReflProb(G4double freq = -1.) const;
+  G4double DiffuseReflProb(G4double freq = -1.) const;
+  G4double SpecularReflProb(G4double freq = -1.) const;
```
kelseymh, line 170: I was completely wrong about this.  These functions are intrinsically phonon-only, and should never be called for charges.  If you want to back out this change, I won't complain.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -168,114 +173,141 @@ void G4CMPDriftBoundaryProcess::DoAbsorption(const G4Track& aTrack,
...
-    // SANITY CHECK:  Does new momentum get back to new velocity?
-    G4ThreeVector vnew = theLattice->MapPtoV_el(GetCurrentValley(),
-						GetLocalDirection(p));
-    RotateToGlobalDirection(vnew);
-    G4cout << " Cross-check new v dir  " << vnew.unit() << G4endl;
+  if (IsElectron())  reflP = DoSpecularElectron(aTrack, aStep);
+  else if (IsHole()) reflP = DoSpecularHole(aTrack, aStep);
+  else {
+    G4Exception("G4CMPDriftBoundaryProcess::DoReflection", "Boundary004",
+                EventMustBeAborted, "Invalid particle for this process.");
```
kelseymh: Should we expand the message text to include the particle type, via `aTrack.GetParticleDefinition()->GetParticleName()`?

kelseymh (changes requested):
I noticed that there are a few places in the new code where we are doing global-to-local rotations.  If you haven't already done so, could you run a test job (SuperSim may be the easiest for this) where you rotated the detector in some weird way, e.g., `/CDMS/Zip/RotXAngle 45 deg` or something else wacky, just to be sure the reflections don't go off the rails?

library/include/G4CMPSurfaceProperty.hh
wlamberson1: I think I should leave this for now and maybe circle back later.

examples/charge/src/ChargeDetectorConstruction.cc
wlamberson1: I have attempted to resolve this in the latest push.

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: I have attempted to resolve this in the latest push.

library/include/G4CMPSurfaceProperty.hh
kelseymh: That works for me.

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -168,114 +173,145 @@ void G4CMPDriftBoundaryProcess::DoAbsorption(const G4Track& aTrack,
...
-						GetLocalDirection(p));
-    RotateToGlobalDirection(vnew);
-    G4cout << " Cross-check new v dir  " << vnew.unit() << G4endl;
+  if (IsElectron())  reflP = DoSpecularElectron(aTrack, aStep);
+  else if (IsHole()) reflP = DoSpecularHole(aTrack, aStep);
+  else {
+    G4ExceptionDescription ed;
+  ed << "Unexpected particle type reaching charge drift boundary reflection: "
+     << aTrack.GetParticleDefinition()->GetParticleName()
+     << " (only electrons and holes are supported)";
```
kelseymh: This looks good!  Could you add "\n" at the beginning of this part of the message?

kelseymh (changes requested):

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: Done!

library/src/G4CMPDriftBoundaryProcess.cc
```diff
@@ -238,7 +238,7 @@ DoSpecularReflection(const G4Track& aTrack, const G4Step& aStep) {
   else if (IsHole()) reflP = DoSpecularHole(aTrack, aStep);
   else {
     G4ExceptionDescription ed;
-  ed << "Unexpected particle type reaching charge drift boundary reflection: "
+  ed << "\nUnexpected particle type reaching charge drift boundary reflection: "
```
kelseymh: This "\n" is redundant and unnecessary.  It's at the very start of the message.  Please move the "\n" at the beginning of the text string at line 243 (in other words, line 243 _should_ look like
```
    << "\n (only electrons and holes are supported)";
```

kelseymh (changes requested):

library/src/G4CMPDriftBoundaryProcess.cc
wlamberson1: Agh, I misunderstood. my bad.

kelseymh (approved):
Combined with the validation plots, this is ready to be merged onto develop.  That won't be done until after the RISQ Workshop (15 June 2026).

PR #84, opened 2026-06-01
G4CMP-619: Softening a few exceptions in Boundary/GeometryUtils from Fatal Exceptions to less extreme types.

relineha:
Updated a few exceptions in GeometryUtils and BoundaryUtils from FatalException to either EventMustBeAborted or JustWarning. 

Did not change ALL of the exceptions in Boundary/GeometryUtils -- there are a handful in GeometryUtils that are specifically applicable to 2D diffusion, that I'd like to keep fatal for now since they shouldn't be a blocker for SCDMS as much as the other ones I have touched here are.

### DIFF
--- a/library/src/G4CMPBoundaryUtils.cc
+++ b/library/src/G4CMPBoundaryUtils.cc
@@ -558,7 +558,7 @@ G4bool G4CMPBoundaryUtils::CheckStepBoundary(const G4Step& aStep,
         //it's in the local coordinate system)
         if (preSolid->Inside(surfacePoint) != kSurface) {
           G4Exception((procName+"::CheckStepBoundary").c_str(),
-                      "Boundary006", FatalException /*EventMustBeAborted*/,
+                      "Boundary006", EventMustBeAborted,
                       "Boundary-limited step cannot find boundary surface point"
                       );
           return false;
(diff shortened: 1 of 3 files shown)


library/src/G4CMPBoundaryUtils.cc
```diff
@@ -558,7 +558,7 @@ G4bool G4CMPBoundaryUtils::CheckStepBoundary(const G4Step& aStep,
         //it's in the local coordinate system)
         if (preSolid->Inside(surfacePoint) != kSurface) {
           G4Exception((procName+"::CheckStepBoundary").c_str(),
-                      "Boundary006", FatalException /*EventMustBeAborted*/,
+                      "Boundary006", EventMustBeAborted,
```
kelseymh, line 561: I still don't entirely understand why this needs to kill the event.  By construction, G4Navigator is "guaranteed" to have ended the step at a volume surface point.  If your calculation of a "better" surface point failed, it feels to me like it should generate a warning, and fall back to the original point.

That said, there's a whole layer of complication with parallel worlds that your calculations don't account for at all.  It's entirely possible that with a parallel world, a step will end right in the middle of a physical (mass world) volume, which is at the boundary of a parallel-world volume.

kelseymh (changes requested):

kelseymh (approved):

PR #85, opened 2026-06-09
G4CMP-517 -- Charge recombination in the bulk, when tracks fall below NTL threshold

kelseymh:
With implementation of [G4CMP-503](https://jira.slac.stanford.edu/browse/G4CMP-503) and [G4CMP-515](https://jira.slac.stanford.edu/browse/G4CMP-515), charges can "bounce" off of crystal surfaces multiple times, losing kinetic energy to NTL phonons along the way. Once they drop below NTL threshold (about 100 ueV or so), there are no processes in the simulation for them to lose energy, so they will never end up in a `fStopButAlive` state to be recombined.

I propose that `G4CMPDriftRecombinationProcess` be modified to query `G4CMPLukeScatteringRate` for the lowest energy threshold, and maybe apply recombination to tracks which fall below that threshold. I'll have to think about how to deal with tracks which are actively bouncing, where the electric field turns them around in a parabolic trajectory, but this seems like the right approach.

See the corresponding JIRA ticket for validation plots showing the recombination-in-bulk, and for a discussion of the logic which has been implemented.


library/include/G4CMPVProcess.hh
```diff
@@ -45,14 +45,18 @@ public:
   virtual void StartTracking(G4Track* track);
   virtual void EndTracking();
 
+  // Compute MFP using scattering rate, without other arguments
+  // NOTE: Subclasses cannot override this; override GetMeanFreePath() below
+  G4double GetMFPfromRate(const G4Track&) const;
```
kelseymh, line 50: Just a note to the reviewers -- this change will trigger recompilation of all the processes in the G4CMP library. However, it's a non-virtual function (on purpose), so the library can be used with user applications without recompiling the latter.

PR #86, opened 2026-06-10
G4 cmp 620 -- RISQ 2026 (quasiparticle) tutorial

relineha:
Adding a new tutorial aimed at exposing RISQ 2026 attendees to quasiparticle tracking and the new phonon/QP physics that's come out of G4CMP-V10. The example is based on the existing quasiparticle example. Most of the work is actually in an analysis macro (ROOT), some small tweaks to the quasiparticle code, and a MUCH more fleshed-out README.md file for that example. No G4CMP _library_ code should have been touched.


examples/quasiparticle/README.md
```diff
@@ -1,71 +1,98 @@
-# Quasiparticle Propagation in G4CMP
+# Tutorial: Quasiparticle Propagation in G4CMP
 
 This example is a demonstration of several features and physics lists that are new to G4CMP as of November 2025. Ryan Linehan, linehan3@fnal.gov made this, so please email him with questions/compliments/complaints.
```
szatschler: Should it be "physics processes" instead of "physics lists"?

examples/quasiparticle/README.md
```diff
@@ -1,71 +1,98 @@
...
-  	* If you intend to have QPs in your simulation, this must be turned on for anything to be accurate.
+## Preliminaries
+
+The preliminary requirements for running this example are the same as those needed for other examples such as the RISQTutorial example from RISQ 2024. For completeness, we will repeat these instructions.
+
+### Installing Geant4 and G4CMP
+We'll start with a reminder that in order to run this example, you'll need to install ROOT and both the geant4 and G4CMP packages. On my machine, each of these has three directories associated with its build: a source directory `XXXXX`, a build directory `XXXXX-build`, and an install directory `XXXXX-install`. On my machine, the base name (`XXXXX`) for my geant4 build is `geant4.10.07.p04`, and the base name for the G4CMP build is `G4CMP_quasiparticle`. NB: the last steps in the process of each of these installations should be to run `make` and `make install` while in the `XXXXX-build` directory, for both geant4 and G4CMP
```
szatschler: The mentioning of G4-10.7 here may be confusing given that we just encouraged all participants to make sure they use the latest Geant4 and G4CMP versions. I don't think it would be an issue to just update the paragraph here to refer to G4-11.4.1. I tested the installation with that version on my end.

examples/quasiparticle/README.md
```diff
@@ -1,71 +1,98 @@
...
 
-In the G4 macro used to run events, these processes can be turned on and off (mostly) at your discretion. An example code block looks like so:
 
+> [!IMPORTANT]
+> To streamline your ability to prep for this tutorial, we recommend installing Geant4 with the cmake flags `-DGEANT4_INSTALL_DATA=ON` and `-DGEANT4_USE_OPENGL_X11=ON`, and to build with C++14. In particular, the OpenGL flag will enable visualization, which we will frequently use. However, if you can successfully run other visualizers like DAWN, those are also perfectly fine.
```
szatschler: Our G4-11.4.1 container version uses C++17.

examples/quasiparticle/README.md
```diff
@@ -1,71 +1,98 @@
...
-#/process/inactivate phononScattering
-#/process/inactivate phononDownconversion
-/process/inactivate phononPolycrystalElasticScattering
-#/process/inactivate qpRecombination
-#/process/inactivate qpRadiatesPhonon
-#/process/inactivate scPairBreaking
-#/process/inactivate qpDiffusion #If QPs can exist, diffusion needs to be active
-#/process/inactivate qpLocalTrapping
-/process/inactivate qpDiffusionTimeStepper
+source /path/to/geant4.10.07.p04-install/bin/geant4.sh
```
szatschler: Update to G4-11.4.1

examples/quasiparticle/README.md
```diff
@@ -1,71 +1,98 @@
...
+source /path/to/G4CMP_quasiparticle-install/share/G4CMP/g4cmp_env.sh
+```
+Now we can make our example. Copy this tutorial's source directory into a new directory -- I like to copy it outside of the whole G4CMP source directory just to avoid confusion and remember that this is its own executable that needs to be made. Moreover, make build and install directories to accompany it:
```
szatschler: Now we can make -> Now we can compile

examples/quasiparticle/README.md
```diff
@@ -1,71 +1,98 @@
...
+cd /path/to/G4CMP_quasiparticle 
+cp -r ./examples/quasiparticle /path/to/
+cd /path/to/
+mkdir quasiparticle-build
+mkdir quasiparticle-install
+```
+Now we head into our build directory and run CMake:
+```
+cd quasiparticle-build
+cmake -DCMAKE_INSTALL_PREFIX=/path/to/quasiparticle-install -DCMAKE_CXX_STANDARD=14 ../quasiparticle/
```
szatschler: Update to C++17. You can also use a relative path here.
DCMAKE_INSTALL_PREFIX=../quasiparticle-install

examples/quasiparticle/README.md
szatschler: The last sentence misses a full stop and I would prefer "Geant4" over "geant4".

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+Rather than being prescriptive about what applications can make best use of this new addition, we propose flavors of problem that this new version can now model better. These include devices whose signal or performance is reasonably dependent on:
+* The locations and energy distribution of quasiparticles produced in an interaction
+* The transport of quasiparticles across regions of varying superconducting gap Δ.
+* Phonon recycling (i.e. QP recombination into phonons that rebreak Cooper pairs)
+Some examples of these include superconducting qubit devices, where diffusion plays a role in governing the population of QPs that can make it to the junction for tunneling, as well as resonators, where different densities of QPs at different points on the resonator may produce variable signals (CITE). We'll look at an example of the latter in our Tutorial Example 2.
```
szatschler: Is (CITE) a placeholder for a paper reference that you want to add here?

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+However, if your application does not need to model these effects, a more limited superconducting response is still handled by the KaplanQP class within G4CMP, and executes significantly faster. This works relatively well for spatially limited devices, but currently ignores QP recombination.
+
+## Tutorial Example 1: Geometry Construction
+
+The goal of this section is to introduce advanced users to new elements of geometry construction that are needed for adopting the new features of G4CMP-V10. We're going to start in the approximately the same place as the original RISQ Tutorial example: by inspecting a moderately complicated superconducting qubit geometry, once again based on the "candlestick" Xmon qubits designed by the McDermott Group at UW-Madison. The following features are on this chip:
```
szatschler: We're going to start in the approximately... -> We're going to start in approximately...

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+* Silicon chip substrate
+* Aluminum ground plane (sibling to the silicon chip in the geometry heirarchy)
+* Aluminum CPW transmission line with wirebond pads (daughter of the ground plane)
+* Aluminum CPW quarter-wave resonators (x6) (daughters of the ground plane)
+* Aluminum Xmon qubits (no junction leads included) (also daughters of the ground plane)
+* Copper mounts for thermalization (sibling to the silicon chip in the geometry heirarchy)
+  
+To see these features, we'll run our first macro, `quasiparticle_geometry_vis.mac`, from our `quasiparticle-build` directory. First, we'll navigate there and boot up G4CMP:
```
szatschler, line 127: If you define a separate build and install directory, then your binary should be in the install area under "quasiparticle-install/bin". This is from where you should start the session.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+```
+cd /path/to/quasiparticle-build
```
szatschler, line 129: cd /path/to/quasiparticle-install

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+./g4cmpQuasiparticle
```
szatschler, line 130: ./bin/g4cmpQuasiparticle

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+#/process/inactivate qpLocalTrapping
+#/process/inactivate qpDiffusionTimeStepper
+```
+Third, let's set the phonon bounces to 100, so that we have ample chance of watching the phonons do interesting things in the film and substrate.
+```
+/g4cmp/phononBounces 100
+```
+Okay with that, let's rerun. 
+```
+./g4cmpQuasiparticle
```
szatschler, line 168: ./bin/g4cmpQuasiparticle

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+```
+Only the first twenty steps are shown, and show a quasi-periodic behavior in which the phonon's volume follows a trajectory like SiliconChip-->World-->SiliconChip-->GroundPlane-->World-->GroundPlane-->SiliconChip. This cadence demonstrates a phonon reflecting off of the vacuum boundaries with the world (the steps after "World" are zero-length turnaround steps) and then transmitting through the SiliconChip-to-Groundplane interface. Here the StepLeng (step length) column for the steps like Step 4, in which in which the last step's NextVolume was the GroundPlane and the current step's NextVolume is World, shows that the phonon is propagating around 0.000122 mm (about 122 nm), which is comparable to the thickness of the ground plane. Together, these things demonstrate that _the ground plane is a physically realized volume in which the phonons can propagate_, which is a new feature of this version of G4CMP.
+
+With that simple phonon example under our belt, let's make things a bit more complicated and see what happens when we turn on quasiparticle physics. Go ahead and exit (`exit`), and let's re-enter our macro, reactivate cooper pair breaking:
```
szatschler: cooper pair breaking -> Cooper pair breaking

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+1. First argument, value `AlLogical`: this is the logical lattice associated with the physical lattice, and the argument `AlLogical` is defined at the beginning of the `QuasiparticleDetectorConstruction.cc` file. This is a package of information from the aluminum CrystalMaps file. 
+2. Second argument, value `dp_polycryElScatMFP_Al`: This is the length scale for energy-independent, elastic phonon scattering off of polycrystalline grains in this thin film volume. 
+3. Third argument, value `dp_scDelta0_Al`: This is the zero-temperature gap value of this thin-film volume.
+4. Fourth argument, value `dp_scTeff_Al`: This is the effective temperature of this thin-film volume. Together with `dp_scDelta0_Al`, this governs the rates of pairbreaking, phonon radiation by QPs, and QP recombination via the calculations/formalism set out in [CITE KAPLAN]
+5. Fifth argument, value `dp_scDn_Al`: This is the normal-state diffusion constant for electrons in the material — this is combined with an energy dependent term via the formalism in [CITE ULLOM] to capture the dependence of a QP’s transport on its energy.
```
szatschler: Replace [CITE ULLOM] with proper reference.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+
+We'll use a new macro this time, `quasiparticle_resonator_targeted.mac`. In this macro, we will simulate high-energy (Debye-energy) phonons in the superconductor as a way of mimicking a fraction of an energy deposition from, say, an optical calibration source, and will store quasiparticle step information for analysis. As we proceed, we'll split this work up into three parts: running the simulation, analyzing the data, and tuning parameters. 
+
+
+### Running the Simulation and Generating Quasiparticle Step Info
+
+Let's go ahead and just run our new macro out of the box:
+```
+./g4cmpQuasiparticle
```
szatschler, line 451: ./bin/g4cmpQuasiparticle

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+/control/execute ../quasiparticle/G4Macros/quasiparticle_resonator_targeted.mac
+```
+It's first set up to launch three events, each with a single 36.9 meV phonon (Al Debye energy, a la [CITE KOZOREZOV]) in the ground plane at Point 1 in the above diagram. This should run for about 3 minutes, and should produce a visualization that looks like the left image below:
```
szatschler: Add proper citation for [CITE KOZOREZOV].

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+mv QuasiparticleStepInformationFile.txt QuasiparticleStepInformationFile_Point1_3evts.txt
+```
+so that we name the file with some useful labels that we can use to remember what went into it. Once we do this, let's go into our macro and edit it to simulate interactions at Point 2, by changing the position definition lines to:
+```
+#/gps/pos/centre -2.65 1.221 5.0001 mm # Point 1, in film
+/gps/pos/centre -1.85 1.221 5.0001 mm # Point 2, in film
+```
+Now rerun the simulation with this amended interaction point, which is dumping our Debye-energy phonon into the central conductor of our CPW resonator at Point 2. 
+```
+./g4cmpQuasiparticle
```
szatschler, line 482: ./bin/g4cmpQuasiparticle

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+/control/execute ../quasiparticle/G4Macros/quasiparticle_resonator_targeted.mac
+```
+This process should take much longer, about 15 minutes, and will give an output as shown on the right half of the above image. (In the meantime, we'll move on and talk about analysis of the output file we've already collected.) Once this finishes, let's go ahead and rename this as well:
+```
+mv QuasiparticleStepInformationFile.txt QuasiparticleStepInformationFile_Point2_3evts.txt
+```
+
 > [!TIP]
-> The _strength_ of the various processes' dependence on the various parameters varies somewhat widely. For example, if the effective temperature parameter `scTeff` is a small (<10%) fraction of the T_{c} for a given superconductor, then QPs will functionally take *forever* for these QPs to recombine. If your execution is taking forever, this could be a culprit.
+> Homework question: why does it make sense that this simulation taking longer? Given the walk-on-spheres technique, can you estimate a naive scaling for how much longer this simulation should take than the simulation for Point 2? Does it agree with the observed difference in execution time?
```
szatschler: why does it make sense that this simulation taking longer? -> Why does it make sense that this simulation takes longer?

examples/quasiparticle/AnalysisTools/quasiparticle_analysis_output.root
szatschler, line 1: Is this a required file or just an example? It is not very big, but we usually do not commit data files to the git history.

szatschler (commented):
I have focused on the README and ran parts of the tutorial. Very nice work! I think this is good to go for the workshop. There are certainly a few things that could be improved at a later stage, but I don't see anything that would hold up this PR.

examples/quasiparticle/AnalysisTools/quasiparticle_analysis.C
kelseymh, line 1: Is this the ROOT macro?  I'd suggest nameing with ".C", rather than ".cc".  Not required, but just a suggestion so it doesn't appear to be part of the build.

examples/quasiparticle/AnalysisTools/quasiparticle_analysis.C
```diff
@@ -0,0 +1,565 @@
+//------------------------------------------------------------------
```
kelseymh, line 6: Can you copy-paste the four-line GNU license block we have at the top of most (supposed to be all...) G4CMP source files?

examples/quasiparticle/AnalysisTools/quasiparticle_analysis.C
```diff
@@ -0,0 +1,565 @@
...
+}
+
+//------------------------------------------------------------------------------
+// Parsing function: bundling the creation of the final event list
+std::vector<Event>
+CreateEventListFromStepFiles(std::string stepFile) {
+  std::vector<Event> output;
+  std::map<std::pair<int,int>,Event> map_runIDEvtID_Event =
+    ParseStepTextFileForEvents(stepFile);  
+  for( std::map<std::pair<int,int>,Event>::iterator it = map_runIDEvtID_Event.begin(); it != map_runIDEvtID_Event.end(); ++it ){
```
kelseymh, line 221: You prefer this to the "new" style?
```
for (const auto& evt: map_runIDeVtID_Event) {
   output.push_back(evt->second);
}
```

examples/quasiparticle/AnalysisTools/quasiparticle_analysis_output.root
kelseymh: I agree with Stefan, here.  The only time data files should be in the repository is if they're being used for a regression test (i.e., new output == expected output).

kelseymh (changes requested):

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
 | `scDn` | Normal-state QP Diffusion Constant | Fifth argument of `G4LatticePhysical` constructor | 6 μm^{2} / ns | `G4CMPQPDiffusion.cc` |
```
BirgitZatschler: I believe markdown uses `<sup></sup>` for superscript.

examples/quasiparticle/G4Macros/quasiparticle_geometry_vis.mac
```diff
@@ -0,0 +1,76 @@
...
+# This just visualizes the geometry that we will
+# use in the quasiparticle example.
+#-------------------------------------
+
+# Initialize the run
+/run/initialize
+
+# We'll use an OpenGL viewer here, and load
+# a specific angle and zoom
+/vis/open OGL
```
BirgitZatschler, line 15: I compiled against the container `G4CMP-base_G4-11.4.1.sif`. The default viewer with OGL is then TOOLSSG_QT_GLES which is missing the right-click menu I'm used to from Geant4-10. Changing `OGL` to `OGLSQt` uses the viewer OpenGLStoredQt which has this right-click menu. This might be a personal preference though.
Same for the other macro.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+   19     1.06    -1.15     4.94     4e-09        0     0.77      6.93 SiliconChip phononScattering
+   20     1.06    -1.14     4.95     4e-09        0   0.0126      6.94 SiliconChip phononScattering
+```
+Only the first twenty steps are shown, and show a quasi-periodic behavior in which the phonon's volume follows a trajectory like SiliconChip-->World-->SiliconChip-->GroundPlane-->World-->GroundPlane-->SiliconChip. This cadence demonstrates a phonon reflecting off of the vacuum boundaries with the world (the steps after "World" are zero-length turnaround steps) and then transmitting through the SiliconChip-to-Groundplane interface. Here the StepLeng (step length) column for the steps like Step 4, in which in which the last step's NextVolume was the GroundPlane and the current step's NextVolume is World, shows that the phonon is propagating around 0.000122 mm (about 122 nm), which is comparable to the thickness of the ground plane. Together, these things demonstrate that _the ground plane is a physically realized volume in which the phonons can propagate_, which is a new feature of this version of G4CMP.
```
BirgitZatschler: the steps like Step 4, in which in which the last step's -> the steps like Step 4, in which the last step's

examples/quasiparticle/AnalysisTools/quasiparticle_analysis_output.root
relineha: Shoot, sorry, I fully missed this but did not intend to add this. I'll remove this.

kelseymh: If you're going to remove it, now would be a very good time.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+Why do we recommend this? The reason gets back to the fact that QPs are only simulated in 2D and cannot currently diffuse in Z: if films vary in Z thickness, a QP may not be able to access superconducting areas if they do not extend to exactly the same Z at which it is diffusing. A couple of example failure modes are diagrammed below.
+
+<img width="1239" height="267" alt="image" src="https://github.com/user-attachments/assets/8dc71869-0af6-4845-ab53-8ab02da9d136" />
+
+To both help us achieve this uniformity in thickness and to improve the accuracy and cleanliness of our code, we also liberally nest volumes within other volumes in our code (as Geant4 encourages). In the code above, the physical volume construction shows that this volume is a daughter of the `log_curve3Empty` volume, which has the same height but just slightly wider radial bounds. In our zoomed-in image, `curve3Empty` is the volume with the gray bounds, and the `curve3Conductor` created here is the thin cyan annulus. We also note that `curve3Empty` itself is a daughter of the larger base aluminum layer of this resonator assembly, and that base aluminum layer is a daughter of the ground plane.
```
BirgitZatschler: this volume is a daughter of the `log_curve3Empty` volume, which has the same height -> This sounds a bit like the daughter and the mother volume share the top and bottom surfaces. This is something G4 discourages to do.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+Sometimes, it may be useful to define geometrical volumes using boolean additions or subtractions of basic Geant4 objects. For example, the Xmon cross volumes (both vacuum and conductor) and the c-shaped coupler on the end of the λ/4 resonator are either single or nested `G4UnionSolids`: 
+
+<img width="650" height="650" alt="image" src="https://github.com/user-attachments/assets/4fd70eed-055c-496b-9522-01b0046ece9c" />
+
+These boolean solids should work reasonably well with the quasiparticle and phonon physics in G4CMP, but...
+1. ...deep nestings of G4UnionSolids to build complicated structures is not wise on performance grounds, and in my experience has occasionally made visualization choke as well. While functionally a triply- or quadruply-nested G4UnionSolid will work okay, consider using a G4MultiUnion if you're going to going to be attempting to link more than a couple basic `G4VSolid` objects (`G4Tubs`, `G4Box`, `G4Trd`, etc.). QP and phonon transport in `G4MultiUnion` objects has been lightly tested and anecdotal evidence points to it giving the correct film response, but this also needs further rigorous exploration, so proceed with caution and skepticism.
```
BirgitZatschler: if you're going to going to be attempting -> if you're going to be attempting

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+> [!TIP]
+> Challenge question: It is possible to define complicated geometries with other construction techniques, which may be more efficient than what we do here. Can you think of a way to define portions of this geometry using the G4ExtrudedSolid class?
+
+
+### Lattices in G4CMP-V10
+
+\With the shapes, materials, and positioning of our superconducting volume placed, we now need to ascribe to it a lattice. While most superconducting films don’t end up as single-crystal lattices without concerted effort during fabrication, we nonetheless ascribe to each superconducting volume a `G4LatticePhysical` object. Let’s look at the lines in `QuasiparticleResonatorAssembly.cc` that do this for `curve3Conductor`:
```
BirgitZatschler: Remove initial slash.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+                             curve3Conductor, AlAlBoundary);
+```
+We see that in this block we've defined four boundaries for this volume: two establishing the Al/vacuum boundary between `curve3Conductor` and `curve3Empty` (i.e. the "sidewall" boundaries to the CPW center conductor), and two establishing the boundaries between `curve3Conductor` and the straight horizontal line 7 conductor object, `shl7Conductor`. 
+
+How should I think about why there should be two? Critically, because both phonons and QPs can now _traverse_ boundaries, we have both the opportunity and oblication to provide information about how those particles should see or behave when impinging upon boundaries from either side. This is required for the Al/Al and Al/Si boundaries used in this work, and while not strictly required for the Al/Vacuum boundaries (since phonons or QPs can't exist in vacuum and aren't expected to impinge on the boundary from the vacuum side), we include the "extra" volume for symmetry and overcaution (and in case we wanted to, for example, turn all vacuum into something like LHe, where phonons could propagate.)
```
BirgitZatschler: oblication -> obligation

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+
+## Tutorial Example 2: Scanning Energy Depositions in a Planar Resonator
+
+We now present a tutorial based on the same geometry that is intended to showcase some of the power of G4CMP-V10's new QP tracking. In this, we will study the time evolution of the quasiparticle density in the top left superconducting resonator on the chip, for energy depositions at two different points: one in the ground plane (Point 1) and one directly within the coplanar waveguide's inner conductor (Point 1).
```
BirgitZatschler: one in the ground plane (Point 1) and one directly within the coplanar waveguide's inner conductor (Point 1) -> one in the ground plane (Point 1) and one directly within the coplanar waveguide's inner conductor (Point 2)

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+However, if your application does not need to model these effects, a more limited superconducting response is still handled by the KaplanQP class within G4CMP, and executes significantly faster. This works relatively well for spatially limited devices, but currently ignores QP recombination.
+
+## Tutorial Example 1: Geometry Construction
+
+The goal of this section is to introduce advanced users to new elements of geometry construction that are needed for adopting the new features of G4CMP-V10. We're going to start in the approximately the same place as the original RISQ Tutorial example: by inspecting a moderately complicated superconducting qubit geometry, once again based on the "candlestick" Xmon qubits designed by the McDermott Group at UW-Madison. The following features are on this chip:
+* Silicon chip substrate
+* Aluminum ground plane (sibling to the silicon chip in the geometry heirarchy)
+* Aluminum CPW transmission line with wirebond pads (daughter of the ground plane)
```
BirgitZatschler: Could you explain CPW?

examples/quasiparticle/README.md
BirgitZatschler: Maybe also that the file size of the output file could be large. Mine is 412 MB.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+The main output from this is a file called `QuasiparticleStepInformationFile.txt`. We'll open that in a sec, but for now let's rename it and rerun, since our next sim will actually take about 15 minutes to run, and it will overwrite this output file if we keep it the same name. After exiting G4CMP's interactive session, we can run
```
BirgitZatschler: About 40 min on my 8 years old Intel® Core™ i7-8565U CPU @ 1.80GHz × 8
And the output is much larger. Mine is 5.4 GB. Maybe add a warning here to prevent people from filling their disk space in case they don't have much.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+                << postStepY_mm << " " << postStepZ_mm << " " << postStepT_ns
+                << " " << postStepEnergy_eV << " " << postStepKinEnergy_eV
+                << " " << nReflections << " " << stepProcess << " "
+                << preStepVolume << " " << postStepVolume << std::endl;
+}
+```
+In this block (which runs after gathering all of the step info that is exported here) we write information to our stepping output file if the particle taking the step is a `BogoliubovQP`. For that QP, we write pre- and post-step information including X, Y, Z, T, E, KE, and volume name variables, as well as the total number of reflections at this step and the process that determines the step. Notably, we do this for _all_ QP steps, without discrimination. As you may have already noticed looking at your output files from the last section, this can produce reasonably large output files even for small numbers of events. For now, we'll accept this, with the recognition that it we can trim these files using additional conditionals in the above block of code to request that the simulation only saves the steps we care about.
```
BirgitZatschler: with the recognition that it we can trim these files -> with the recognition that we can trim these files

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
 
-## Geometry Description
+> [!CAUTION]
+> Running far more events in our `quasiparticle_resonator_targeted.mac` than what we've already set up without further conditionals on what is saved may quickly fill up your disk. Proceed with caution.
```
BirgitZatschler: I suggest to add something like (if you know/if this is reasonable): expect about X GB per event.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+With this information, let's go ahead and run our analysis macro using ROOT. We'll start up an interactive ROOT session and run
+```
+root -l
+.L ../quasiparticle/AnalysisTools/quasiparticle_analysis.cc
+run_quasiparticle_analysis("/path/to/QuasiparticleStepInformationFile_Point1_3evts.txt","/path/to/QuasiparticleStepInformationFile_Point2_3evts.txt")
+```
+This will take some time to run -- it reads the step files, which are between a few hundred MB and a few GB, and builds simple event structures out of them. It then analyzes those events, plotting basic information like quasiparticle creation/destruction times and locations, all step locations, and finally the time-averaged quasiparticle occupations of our resonator structures. We'll first explore some basic cross-check plots, which is an important step in any analysis to make sure that our desired results make sense.
```
BirgitZatschler: I would not call these "step files". People may confuse these with CAD .step files. Why not just "output files"?

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
 
-## Geometry Description
+> [!CAUTION]
+> Running far more events in our `quasiparticle_resonator_targeted.mac` than what we've already set up without further conditionals on what is saved may quickly fill up your disk. Proceed with caution.
 
-In this example, the geometry is a superconducting qubit chip with six "candlestick" qubits adjacent to a superconducting feedline. Phonons may transmit from the silicon chip into this superconducting layer (as well as the copper housing meant for thermalization), and may produce quasiparticles that diffuse in 2D around the layer.
+With this information, let's go ahead and run our analysis macro using ROOT. We'll start up an interactive ROOT session and run
+```
+root -l
+.L ../quasiparticle/AnalysisTools/quasiparticle_analysis.cc
```
BirgitZatschler: We don't compile the script? According to my experience compiled C++/ROOT code used to be much faster than just loading the script. This might be a thing of the past though. In any case, compilation would fail on my side.

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+1. As QPs radiate off phonons, those phonons may leave the superconducting film, bounce around the chip to different, more distant locations in the chip and re-break pairs in the film there.
+2. QPs that recombine in the vicinity of the initial Debye-energy phonon may produce near 2Δ phonons that can then go pairbreak elsewhere as well.
 
-## Primary Event
+So okay, this seems sensible. For XZ and YZ, we do see that all of our QP creation points occur in the thin film on the top of the 400μm chip, and there are no other QP steps taken elsewhere (as expected).
 
-## Execution and Output
+Let's move on to looking cross-checking our plots of _all_ QP steps. Here, now focusing entirely on XY because our XZ and YZ plots all (thankfully) look the same, we find the following plot for Point 1.
```
BirgitZatschler: move on to looking cross-checking our plots -> move on to cross-checking our plots

examples/quasiparticle/README.md
```diff
@@ -78,35 +105,572 @@ On top of the "simple" phonon parameters needed to be defined for any material,
...
+
+<img width="400" height="142" alt="image" src="https://github.com/user-attachments/assets/8430e220-d72a-49cb-8411-fac6e59a384b" />
+
+Where D_n is the normal-state diffusion coefficient just above the transition temperature. In G4CMP, our value for this is about 6 um^2/ns. The dependence on the QP energy means that for most QPs that have cooled, the diffusion coefficient will be a factor of a few lower than this D_n out front, so we'll guess for now that the diffusion coefficient is about 2 um^2/ns. Then, we can calculate what average Δt we may expect for a given diffused Δx using (Δx)^2/2D. For our numbers here, using Δx=2400 um and D=2 um^2/ns, we find an expected time to diffuse of about 1.1 ms. For a back of the envelope, this is reasonably close to what we see, which is a nice confirmation.
```
BirgitZatschler: You can use `<sup></sup>` and `<sub></sub>` for superscript and subscript in markdown.

kelseymh: Just FYI, Birgit, whatever characters you put in your comment vanished (maybe because the comments are in Markdown?).

BirgitZatschler: *facepalm*
Thanks for noticing, Mike. Fixed.

examples/quasiparticle/README.md
kelseymh: You may be right, Birgit.  The README.md file evolved (badly!) from our original plain-text README.  I didn't know that Markdown supported actual HTML tags.  That being said, I have a small prejudice to keeping it as close to "plain text" as possible, so users (like me) can read it using `more` in their terminal.  But I'm not going to fight it if y'all choose to change it.

BirgitZatschler: I'm fine with both.

examples/quasiparticle/README.md
kelseymh: Yes.  "Geant4" (capital G) is the preferred notation by the Geant4 Collaboration for plain text applications.  In LaTex, `{\sc Geant4}` is preferred.

relineha: This is now fixed.

examples/quasiparticle/README.md
relineha: I've rearranged the tutorial so that we emphasize the recommended installation instructions at the beginning, and then say, "if you want to build yourself, then here are some tips." This puts less of a focus on the installation instructions here and more on the stuff you've been working on, which I think is better. Since I've not yet attempted to build in this manual way with C++17, I think here I may just remove the reference to the C++ version to avoid confusing people, if that is acceptable.

examples/quasiparticle/README.md
relineha: Good catch, thanks.

examples/quasiparticle/README.md
relineha: Done. Thanks!

examples/quasiparticle/README.md
relineha: Updating C++ version, but will keep the absolute path here.

examples/quasiparticle/README.md
relineha: Yeah it was, but after asking the guy who has done that study, it seems like they have not yet published it. I'm going to remove.

examples/quasiparticle/README.md
relineha: Good catch. Thanks.

examples/quasiparticle/README.md
relineha: For me, the binary always lands in the build directory _and_ the install/bin directory. I'll leave as-is since it's simpler but will comment that if it doesn't show up in the build directory, users can look in their respective install directories.

examples/quasiparticle/README.md
relineha: See above comment about the executable showing up in the build directory

examples/quasiparticle/README.md
relineha: See above comment about the executable showing up in the build directory

examples/quasiparticle/README.md
relineha: See above comment about the executable showing up in the build directory

examples/quasiparticle/README.md
relineha: Done.

examples/quasiparticle/README.md
relineha: Done.

examples/quasiparticle/README.md
relineha: See above comment about the executable showing up in the build directory

examples/quasiparticle/README.md
relineha: Done.

examples/quasiparticle/README.md
relineha: See above comment about the executable showing up in the build directory

examples/quasiparticle/README.md
relineha: Good catch. Thanks!

examples/quasiparticle/AnalysisTools/quasiparticle_analysis.C
relineha: Ope, good point. Will rename.

examples/quasiparticle/README.md
relineha: Technically, yes. Good catch. Fixed.

examples/quasiparticle/AnalysisTools/quasiparticle_analysis.C
relineha: Yes. I absolutely hate the "new" style because I have no idea what is being stored/iterated over if I look at the loop. Call me old fashioned but in my own code I never use this... (I know the new style shortens and cleans up the line, but for me it removes critical information.)

examples/quasiparticle/README.md
relineha: Hmmmmmmmm. Will think about this.

examples/quasiparticle/README.md
relineha: Good catch. Thanks!

examples/quasiparticle/README.md
relineha: I did not know that, interesting. And yes, they share the same top and bottom surfaces. Unfortunately, this is liberally used throughout the example. I guess I wouldn't know how to make a flush 2D plane using parent-child relationships without doing this, though. I probably don't have time to go through and change this around before the tutorial, but do you have any suggestions on how not to do this in the future? Maybe I can mention that verbally or add it as a caution flag in the tutorial.

examples/quasiparticle/README.md
relineha: Good catch, thanks!

examples/quasiparticle/README.md
relineha: Done. Good call!

examples/quasiparticle/README.md
relineha: Good call. Now called "stepping output" files.

examples/quasiparticle/README.md
relineha: This is mostly a habit of mine -- I tend just to load. Can try to see if I can compile, but may check this after other things are tidied up.

examples/quasiparticle/README.md
relineha: Done.

examples/quasiparticle/README.md
relineha: Done.

examples/quasiparticle/README.md
relineha: Done. Coplanar waveguide

examples/quasiparticle/README.md
relineha: Yeah execution times may vary, but hopefully the output filesize won't vary. After changing things up, I've brought this filesize down to about 3G, which I think is the lower bound that I can realistically get away with and still make this tutorial clear.

examples/quasiparticle/README.md
relineha: Done.

examples/quasiparticle/README.md
relineha: Done. Thanks!

examples/quasiparticle/README.md
relineha: Yeah that seems about right -- have now tweaked parameters to reduce filesize so this should now be about 150 or so MB.

examples/quasiparticle/README.md
relineha: Good catch. Thanks!

examples/quasiparticle/AnalysisTools/quasiparticle_analysis.C
relineha: Done.

examples/quasiparticle/README.md
relineha: Think I prefer to leave as plain-text for now, but can revisit.

examples/quasiparticle/G4Macros/quasiparticle_geometry_vis.mac
relineha: I will keep OGL as the "base" value here but will add this as a suggestion for people.

examples/quasiparticle/README.md
relineha: Fixed.

examples/quasiparticle/README.md
relineha: Nvm, swapped over -- there's enough other html in this file (images, etc.) that this is a small offender and makes the readme look neater if implemented.

examples/quasiparticle/README.md
kelseymh: I'd add a caution in either the README, or better, as a one-line comment in the code.

We can talk about why separately (it has to do with floating point issues, safety, overlaps, etc.).

examples/quasiparticle/README.md
kelseymh: At the ROOT prompt, it can be handled either way.  I *think* (because I never actually do it) that .X compiles on the fly, rather than using the interpreter.

I'm also hoping that your ROOT macros will be named .C, and not .cc; I think you've already renamed the file?

examples/quasiparticle/AnalysisTools/quasiparticle_analysis.C
kelseymh, line 1: Will this get renamed to .C?

kelseymh (changes requested):

examples/quasiparticle/AnalysisTools/quasiparticle_analysis.C
relineha: Yes - it has been now.

examples/quasiparticle/README.md
kelseymh: That's true, because the "install" step copies the _results_ of building from the build directory.  However, in general the build directory is not required to be kept (and is explicitly not kept when building LMod modules or building the container).  Good practice is that anything required for execution should be copied to the install directory; CMake provides many hooks for that, and standard directory names (like "share/").

examples/quasiparticle/README.md
kelseymh: Change .cc to .C to match the file rename.

kelseymh (changes requested):

kelseymh (approved):

PR #87, opened 2026-06-11
G4CMP-626: RISQ2 Tutorial 1

jwi3:
Creating a simple Geant4-G4CMP adaptation tutorial to present at RISQ 2026 in Madison, WI.


examples/g4beginner/B1_working_Nature_device.zip
szatschler, line 1: This file should be removed.

examples/g4beginner/backup/event_summary_100um.png
szatschler, line 1: Should be removed if not needed. I guess this would apply to all files under "backup".

examples/g4beginner/backup/g4cmp_electrons_100um.root
szatschler, line 1: Please do not commit data files to the code repository. All these should be removed. We should also make sure that they are removed from the git history.

examples/g4beginner/backup/g4cmp_electrons_300um.root
szatschler, line 1: See above.

examples/g4beginner/backup/g4cmp_phonons_100um.root
szatschler, line 1: See above.

examples/g4beginner/backup/g4cmp_phonons_300um.root
szatschler, line 1: See above.

examples/g4beginner/backup/g4beginner.cc
szatschler, line 1: Remove?

examples/g4beginner/G4Macros/init_vis.mac
szatschler, line 1: Where is the vis.mac file?

examples/g4beginner/backup/g4beginner.cc
```diff
@@ -0,0 +1,126 @@
...
+// * technical work of the GEANT4 collaboration.                      *
+// * By using,  copying,  modifying or  distributing the software (or *
+// * any work based  on the software)  you  agree  to acknowledge its *
+// * use  in  resulting  scientific  publications,  and indicate your *
+// * acceptance of all terms of the Geant4 Software license.          *
+// ********************************************************************
+//
+//
+/// \file exampleB1.cc
+/// \brief Main program of the B1 example
```
kelseymh: Please fix these Doxygen markers to refer to this code, not G4.

examples/g4beginner/backup/g4cmp_electrons_100um.root
kelseymh, line 1: Are these ROOT files necessary?  We do not normally include binary data files in the repository, and these are substantial in size.  If they aren't necessary, I will have to use BFG to remove them permanently.

examples/g4beginner/backup/event_summary_100um.png
kelseymh, line 1: Are these figures in backup/ duplicates of the ones in figures/?  Is there a good reason to have both?

examples/g4beginner/include/chargeConfigManager.hh
kelseymh, line 1: Would it be hard to restore the header comments block that was removed when you copied this from examples/charge?

examples/g4beginner/src/B1DetectorConstruction.cc
```diff
@@ -0,0 +1,367 @@
...
+#include "G4Trd.hh"
+#include "G4LogicalVolume.hh"
+#include "G4PVPlacement.hh"
+#include "G4SystemOfUnits.hh"
+#include "G4VisAttributes.hh"
+#include "G4Colour.hh"
+#include "G4Tubs.hh"
+
+// Begin import G4CMP-related header files
+#include "G4MaterialPropertiestable.hh"
```
szatschler: This should be "G4MaterialPropertiesTable.hh" (with a capital T in Table). Otherwise it doesn't compile for me using cmake.

examples/g4beginner/include/B1Analysis.hh
```diff
@@ -0,0 +1,37 @@
...
+//
+/// \file B4Analysis.hh
+/// \brief Selection of the analysis technology
+
+#ifndef B4Analysis_h
+#define B4Analysis_h 1
+
+//#include "g4root.hh"
+#include "g4csv.hh"
+//#include "g4xml.hh"
```
kelseymh: Do the ROOT macros (.C files) assume that `g4froot.hh` was used here, to produce the files that they read in?

examples/g4beginner/include/B1AnalysisConfig.hh
```diff
@@ -0,0 +1,31 @@
...
+
+    G4bool IsElectronMode() const;
+    G4bool IsPhononMode() const;
+
+    G4String GetOutputFileName() const;
+  private:
+    B1AnalysisConfig();
+    ~B1AnalysisConfig();
+
+    static B1AnalysisConfig* fInstance;
```
kelseymh, line 25: Is B1AnalysisConfig guaranteed to be instantiated only in the Master thread?  If not, should this be G4ThreadLocal?  Or is this a direct copy from G4 (in which case we should keep it the way they do).

examples/g4beginner/src/B1AnalysisConfig.cc
```diff
@@ -0,0 +1,65 @@
+#include "B1AnalysisConfig.hh"
+
+#include "G4GenericMessenger.hh"
+#include "G4SystemOfUnits.hh"
+#include "G4ios.hh"
+
+B1AnalysisConfig* B1AnalysisConfig::fInstance = nullptr;
```
kelseymh, line 7: See my comment in B1AnalysisConfig.hh: if this is not restricted to Master-thread only, this should be G4ThreadLocal.

examples/g4beginner/src/B1PrimaryGeneratorAction.cc
```diff
@@ -0,0 +1,92 @@
...
+  G4ParticleTable* particleTable = G4ParticleTable::GetParticleTable();
+  G4String particleName;
+  G4ParticleDefinition* particle
+    = particleTable->FindParticle(particleName="gamma");
+  fParticleGun->SetParticleDefinition(particle);
+  fParticleGun->SetParticlePosition(G4ThreeVector(-195*mm,0,0));
+  fParticleGun->SetParticleMomentumDirection(G4ThreeVector(1, 0, 0));
+  //fParticleGun->SetParticleEnergy(0.1*MeV);
+
+ */
```
kelseymh: Is all of this commented out code relevant?  Is it something that users should reasonably expect to uncomment?  If so, please provide a description of when they should use it.  If it's just "left over" from the Geant4 version, please remove it.

examples/g4beginner/g4beginner.cc
```diff
@@ -0,0 +1,129 @@
...
+  //
+  if ( ! ui ) { 
+    // batch mode
+    G4String command = "/control/execute ";
+    G4String fileName = argv[1];
+    UImanager->ApplyCommand(command+fileName);
+  }
+  else { 
+    // interactive mode
+    UImanager->ApplyCommand("/control/execute init_vis.mac");
```
szatschler: This line fails if the executable is not run from the directory containing the init_vis.mac file. I would propose to remove it and add to the tutorial to run the macro snippet in the GUI.

examples/g4beginner/src/B1PrimaryGeneratorAction.cc
```diff
@@ -0,0 +1,92 @@
...
+{
+  G4ParticleTable* particleTable = G4ParticleTable::GetParticleTable();
+  G4ParticleDefinition* particle = particleTable->FindParticle("e-");
+
+  fParticleGun->SetParticleDefinition(particle);
+  fParticleGun->SetParticleEnergy(100*keV);
+
+  // Start above the Si slab and aim downward into it
+  fParticleGun->SetParticlePosition(G4ThreeVector(0., 0., -3.2*mm));
+  fParticleGun->SetParticleMomentumDirection(G4ThreeVector(0., 0., -1.));
```
kelseymh: These lines will override any macro settings of `/gun/particle`, `/gun/energy`, etc., from the user, which would be confusing.  I'd suggest putting _these_ in the constructor (replacing the big commented out block).

examples/g4beginner/src/B1SteppingAction.cc
```diff
@@ -0,0 +1,205 @@
...
+  G4int speciesID = -1;
+
+  auto config = B1AnalysisConfig::Instance();
+
+  if (config->IsElectronMode()) {
+    if (particleName == "G4CMPDriftElectron") {
+      acceptTrack = true;
+      speciesID = 0;
+    }
+  }
```
kelseymh, line 115: No love for the holes, eh?

szatschler (commented):
Stopping here for now. Please try to address the comments made so far. The most pressing are:
- Restrict number of threads
- Add missing vis.mac
- Remove unnecessary files (in particular data files)

examples/g4beginner/CMakeLists.txt
```diff
@@ -0,0 +1,97 @@
+#----------------------------------------------------------------------------
+# Setup the project
+cmake_minimum_required(VERSION 3.8...3.18)
```
kelseymh: I'm not a CMake expert.  Is this version range prohibiting newer versions of CMake?  My Mac has 4.2.3, and I think our containers have 3.24.

examples/g4beginner/src/B1DetectorConstruction.cc
jwi3: Ugh. I changed this last night and Jesse must have changed it back for some reason I can't imagine. I don't know how it even compiles on his machine.

examples/g4beginner/G4Macros/init_vis.mac
jwi3: Jesse removed it. `init_vis.mac` is functionally the same thing, though, no? I suppose it's no harm to have both.

examples/g4beginner/backup/g4beginner.cc
jwi3: Yeah Jesse reinstated the `backup` folder last night after I'd removed it. I don't know what purpose it serves.

ChangeHistory
```diff
@@ -6,23 +6,8 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
...
-2026-05-20  G4CMP-576 : First step of G4CMPEqEMField uses base EqOfMotion for 
-                        non G4CMP carrier and holes.
-2026-05-18  G4CMP-598 : Make G4CMPEnergyPartition respect nParticlesMinimum
-
-2026-05-06  G4CMP-592 : Replace FatalExceptions in QPDiffusion with warnings
-2026-04-28  G4CMP-496 : Test for zero energy before generating charge pairs.
-2026-04-28  G4CMP-599 : Add track ID and name to G4CMPTrackLimiter output.
-2026-04-22  G4CMP-597 : Make fileCmd optional and provide default value.
+2026-06-10  G4CMP-626 : Add 'g4beginner' tutorial/example.
+2026-04-22	G4CMP-597 : Make fileCmd optional and provide default value.
```
kelseymh: We will want to merge in the HEAD of develop, to pickup all the updates between 22 April and now.

.gitignore
```diff
@@ -17,10 +17,10 @@ doxygen/
 .g4cmp-version
 CMakeLists.txt.user*
 epositions.txt
-*.png
+# *.png
```
kelseymh: I would prefer to keep this ignore here.

You can create a specific `examples/g4beginner/.gitignore` file to override this one.  Put in just one line:
`!*.png`
This wil "unignore" for that directory.  Even better would be to put one of those in the specific .../backup/ and .../figures/ directories instead of the main.  That way users don't accidentally commit the .png files they just created.

examples/g4beginner/B1_working_Nature_device.zip
jwi3: Jesse specifically calls out this zip file in the tutorial. Could we host it on the Confluence page and link to it in the Readme instead of keeping it in the repo?

kelseymh (changes requested):

examples/g4beginner/src/B1DetectorConstruction.cc
kelseymh: Some filesystems (especially _old_ Macs) are case insensitive, and treat lowercase and uppercase equivalently.  Back in the day, you couldn't rename a file by changing case -- you had to rename to temp or something, then rename back to the uppercase version.

examples/g4beginner/B1_working_Nature_device.zip
szatschler: Yes, that's fine with me.

examples/g4beginner/g4beginner.cc
jwi3: This line is copied directly from example B1 in Geant. All the Geant examples implicitly assume you're running the executable from the build/install directory, because they all copy their macro files into that directory in their CMakeLists.txt. I think this would be a logical change to make, but since it's in the original example I'm reluctant to change it.

examples/g4beginner/include/B1Analysis.hh
jwi3: This file isn't used at all. It's been deleted.

examples/g4beginner/backup/event_summary_100um.png
jwi3: Also deleted.

examples/g4beginner/g4beginner.cc
szatschler: I see. But the CMakeLists.txt only copies the macros to the build directory, not the install target. So the execution works when the executable is called from the build directory, but not from the install directory.

examples/g4beginner/G4Macros/init_vis.mac
kelseymh: The problem is that init_vis.mac ends with the line
```
/control/execute vis.mac
```
which thereby _must be in the user's current working directory_.  I really don't think I would ever have something like this, or do it in the main().  If the user doesn't give a macro on the command line, just give them the PreInit> prompt or the Qt GUI or whatever.

kelseymh (changes requested):

jwi3:
The macro directory issue is handled by creating a compiler definition pointing to the `G4Macros` directory and then adding that value to Geant's macro search path.

examples/g4beginner/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+# Setup the project
+cmake_minimum_required(VERSION 3.10)
+if(${CMAKE_VERSION} VERSION_LESS 3.12)
+  cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
+endif()
+set(PROJ_NAME g4beginner)
+project(g4beginner)
+
+if (NOT DEFINED MACRO_DIR)
+  set(MACRO_DIR ${PROJECT_SOURCE_DIR}/G4Macros)
```
kelseymh: This, and all of it's usage below, also needs to be handled in the GNUmakefile.  That needs to happen _after_ the `include g4cmp.gmk` directive, because it has to be done after Geant4's `binmake.gmk` is processed.

examples/g4beginner/g4beginner.cc
```diff
@@ -0,0 +1,132 @@
...
+  // Initialize visualization
+  //
+  G4VisManager* visManager = new G4VisExecutive;
+  // G4VisExecutive can take a verbosity argument - see /vis/verbose guidance.
+  // G4VisManager* visManager = new G4VisExecutive("Quiet");
+  visManager->Initialize();
+
+  // Get the pointer to the User Interface manager
+  G4UImanager* UImanager = G4UImanager::GetUIpointer();
+  const G4String macro_dir = MACRO_DIR;
```
kelseymh: I'm not a huge fan of this method, but we can work on changing it in the future.  In the mean time, please put in a clear comment saying that MACRO_DIR is defined at build time in CMakeLists.txt and GNUmakefile (see my other comment), to point to the source directory of the example.

examples/g4beginner/g4beginner.cc
kelseymh: I thought that if the CMake `-DINSTALL_EXAMPLES=ON` option was set (the default is OFF), then the examples would be copied to install.

examples/g4beginner/GNUmakefile
```diff
@@ -0,0 +1,5 @@
+# G4CMP/g4beginner/GNUmakefile -- for building Geant4-G4CMP conversion demo
+# $Id$
+
+G4CMP_NAME := g4beginner
+include $(G4CMPINSTALL)/g4cmp.gmk
```
kelseymh, line 5: Appropriate Makefile code for setting MACRO_DIR needs to be added after this line.

kelseymh (changes requested):

examples/g4beginner/GNUmakefile
jwi3: Consolidating the above points into one conversation. This is a really frustrating issue. Geant4 examples regularly expect that users will only have macros available in their build directory. None jump through the hoops I'm jumping through here to accommodate all the different potential ways a user will run the code. I would rather just go back to the original method, where we just copy the files into the build directory and forget about it.

kelseymh: I fully agree.  I'm not sure we need to do more than that.  All the G4 examples are build around the assumption that the user copies the directory tree, and runs CMake (or Make) on it as a separate application/installation.

The part I'm uncomfortable about is doing it just halfway.  If you want to roll things back, please do.

My personal preference would be to not even have these "init_vis.mac" and "vis.mac" macros.  Those are something inherited from Geant4, and I think they're a big mistake.  If I don't specify a macro to run, I don't want something being done behind my back.

szatschler: I agree with that point. I'm okay with having a vis.mac file SOMEWHERE that could be run with `/control/execute` from the GUI (or passed to the executable).

jwi3: So Jesse never made any changes to `exampleB1.cc` (renamed to `g4beginner.cc`), `init_vis.mac`, or `vis.mac` from the original G4 example. I just rolled back to that setup. It's a stupid way to do it, but the further we get away from the original example, the more it'll confuse people.

Another push incoming.

examples/g4beginner/src/B1DetectorConstruction.cc
```diff
@@ -0,0 +1,367 @@
...
+    
+    /*
+    
+   //Argon gas definition taken from NIST, then change density (https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/Appendix/materialNames.html)
+   G4double density_Ar = 0.0017857*g/cm3;
+   
+   G4Material* Ar_mat = nist->BuildMaterialWithNewDensity("Ar_newDens","G4_Ar",density_Ar);
+
+
+  //borosilicate (Pyrex) definition, taken from material compendium Borosilicate glass
```
kelseymh: We do *not* need to change this now, but the built in `G4NistManager` has a material `G4_Pyrex_Glass` which should take care of all of this automatically.

kelseymh (changes requested):

examples/g4beginner/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+```python
+  G4CMPConfigManager::Instance();
+```
+<section> 
+  <h2>Build the Tutorial Example</h2>
+  <p></p>
+</section>
+
+Now we're ready to see what the tutorial application can do. Create a build directory within the source directory, descend into it, and compile the program with
```
BirgitZatschler, line 107: Creating the build directory within the source directory seems unusual to me. Normally, you would want to keep the source, build and install directories separate.

kelseymh: I agree, and I really don't like it, but I've also seen this with some of the G4 examples, as well as some of the SuperSim packages :-(  Jake added "build/" and "install/" to .gitignore in order to protect things.

examples/g4beginner/GNUmakefile
kelseymh: Yeah, I saw all of the "B1" filename TLAs.  I guess it's to make the point about "start with your favorite G4 example, and here's how to add G4CMP to it.  Now you can do it with your own application."

examples/g4beginner/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+# You can set WITH_GEANT4_UIVIS to OFF via the command line or ccmake/cmake-gui
+# to build a batch mode only executable
+#
+option(WITH_GEANT4_UIVIS "Build example with Geant4 UI and Vis drivers" ON)
+if(WITH_GEANT4_UIVIS)
+  find_package(Geant4 REQUIRED ui_all vis_all)
+else()
+  find_package(Geant4 REQUIRED)
+endif()
+find_package(G4CMP REQUIRED)
```
BirgitZatschler: This doesn't configure on my side. I need to add the G4CMPINSTALL environment variable to CMAKE_PREFIX_PATH first:
```
set(CMAKE_PREFIX_PATH $ENV{G4CMPINSTALL}/../..)
find_package(G4CMP REQUIRED)
```

examples/g4beginner/README.md
```diff
@@ -0,0 +1,253 @@
...
+make
+```
+
+<img src="figures/Zoomed_geometry_visualization.png" width="50%">
+
+You can inspect the geometry setup by running the program with the original visualization macro
+```bash
+./g4beginner init_vis.mac
+```
+The geometry used here is meant to mimic the setup studied by Vepsäläinen et al.: a copper source disk above a superconducting aluminum structure on a silicon substrate (200 nm of Al on 280 μm of Si) with lateral dimensions 5 mm x 5 mm. The source is $^{64}\mathrm{Cu}$ with an activity of $6.12\,\mu\mathrm{Ci}$. It emits $\beta^+$ particles (653 keV; 17.49%), $\beta^-$ particles (580 keV; 38.5%), and \(\gamma\) rays (1346 keV; 0.472%), with emission taken to be isotropic from the Cu volume. The original application included all of this physics, including $\beta$ emission sampled from a triangular energy distribution. For the purposes of this tutorial, however, we simplify things considerably: the particle gun fires only $\beta^-$ particles, at a fixed energy of 100 keV and from a fixed distance of about $350\,\mu\mathrm{m}$. As a visual aid, Fig. 1 from the paper is reproduced below:
```
BirgitZatschler: I suggest to add for beta^- and beta+ that these are the maximum energies, e.g. E_max = 653 keV, so no one thinks these might be fixed energies.

examples/g4beginner/CMakeLists.txt
szatschler: Confirmed on my end for the G4-11.4.1 developer container.

examples/g4beginner/README.md
```diff
@@ -0,0 +1,253 @@
...
+cd build
+cmake ../
+make
+```
+
+<img src="figures/Zoomed_geometry_visualization.png" width="50%">
+
+You can inspect the geometry setup by running the program with the original visualization macro
+```bash
+./g4beginner init_vis.mac
```
BirgitZatschler: This runs without opening any visualization for me. At the end it seems to crash with:
```
RunManagerKernel is deleted. Good bye :)
RunManager is deleted.
Segmentation fault (core dumped)
```
Just running `./g4beginner` opens the Qt interface on my side and I can see the visualization directly without executing any macro. Closing the window results in the same crash as above. Running the other macros end with the same crash as well.

examples/g4beginner/README.md
```diff
@@ -0,0 +1,253 @@
...
+./g4beginner g4cmp_electrons_100um.mac > g4cmp_electrons_100um.out
+```
+where the `g4cmp_electrons_100um.mac` macro is:
+```python
+/run/initialize
+/control/verbose 2
+/run/verbose 2
+/event/verbose 0
+/tracking/verbose 1
+
```
BirgitZatschler, line 146: This is not up to date anymore. In the current version of this macro we have additional lines here:
```
/gun/particle e-
/gun/energy 100 keV
/gun/position 0 0 3.2 mm
/gun/direction 0 0 -1
```

examples/g4beginner/README.md
```diff
@@ -0,0 +1,253 @@
...
+In the third block, note that phonons are omitted (or killed) in this simulation. This is done simply to speed up the collection of electron statistics. The fourth block controls the mean free paths for electron and hole trapping. We will vary these momentarily to see how they affect the charge dynamics. Finally, we use 100 shots here to obtain results quickly, with the tradeoff of poorer statistics. For the plots shown below, I increased the number of shots to 1000; on my machine, that took about 5 minutes to run.
+
+Once the code has finished running, we will use ROOT to launch a plotting script called `g4cmp_electron_interface_hits.C`,
+
+```bash
+cp g4cmp_electrons.root g4cmp_electrons_100um.root
+root -l 'g4cmp_electrons_interface_hits.C("e-",100)'
```
BirgitZatschler: I see a warning/error when running the ROOT script:
```
Apptainer> root -l 'g4cmp_electrons_interface_hits.C("e-",100)'
root [0] 
Processing g4cmp_electrons_interface_hits.C("e-",100)...
Error in <HandleInterpreterException>: Trying to dereference null pointer or trying to call routine taking non-null arguments
Execution of your code was aborted.
In file included from input_line_8:1:
/home/birgit/soft/apptainer/G4CMP/G4CMP-626-g4beginner_build/g4cmp_electrons_interface_hits.C:53:3: warning: null passed to a callee that requires a non-null argument [-Wnonnull]
  hKE->Scale(1.0 / nShots);
  ^~~

```
I only get a white canvas. Nothing happens after that. This is certainly because there are no hits in my .root output.

examples/g4beginner/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+In the third block, note that phonons are omitted (or killed) in this simulation. This is done simply to speed up the collection of electron statistics. The fourth block controls the mean free paths for electron and hole trapping. We will vary these momentarily to see how they affect the charge dynamics. Finally, we use 100 shots here to obtain results quickly, with the tradeoff of poorer statistics. For the plots shown below, I increased the number of shots to 1000; on my machine, that took about 5 minutes to run.
```
BirgitZatschler: After increasing run/beamOn to 1000, the simulations runs just a few seconds on my device and the ROOT file has no hits:
```
root [2] InterfaceHits->Show(0)
Error in <TTree::Show()>: Cannot read entry 0 (entry does not exist)
root [3] EventSummary->Show(0)
======> EVENT:0
 eventID         = 868
 nInSi           = 0
 nReached        = 0
 nTrapped        = 0
 reachFraction   = 0
root [4] PhononEventSummary->Show(0)
======> EVENT:0
 eventID         = 868
 nPhononsInterfaceWeighted = 0
 ePhononsInterfaceWeighted_eV = 0
 nPhononsAbove2DeltaWeighted = 0
 ePhononsAbove2DeltaWeighted_eV = 0
```
I think something had been changed for running this macro, which I either missed or wasn't mentioned.

examples/g4beginner/src/B1DetectorConstruction.cc
jwi3: I just deleted all of it. It's never used in the example.

examples/g4beginner/README.md
jwi3: I'll admit: I do this all the time because I'm lazy and then immediately regret it when I want to run multiple builds side-by-side. It's terrible practice but it's deeply ingrained in many of the G4 examples, so I'm just "going along to get along" here.

examples/g4beginner/README.md
jwi3: If you don't provide a macro as an argument to the command, it'll run `init_vis.mac` (and consequently `vis.mac` as well) by default to pop the vis window (I don't think many users are comfortable with the command-line based tcsh-style interface). I briefly tried to chase down the segfault but didn't get very far in ~20 minutes of looking at it. G4CMP end-of-run segfaults are the bane of my existence: the comments on [G4CMP-560](https://jira.slac.stanford.edu/browse/G4CMP-560) show me slowly losing my mind over one we only ever kind of figured out. I don't have much hope that I'll solve this one before Monday.

examples/g4beginner/README.md
jwi3: Done; I called them endpoint energies since that's the terminology I'm familiar with, but I'm willing to be overruled on that.

examples/g4beginner/README.md
jwi3: Fixed. Oversight on my part.

examples/g4beginner/README.md
jwi3: This SHOULD be fixed now that I've fixed the gun position in the macros (it was 3.2 mm before, but should have been -3.2 mm). If it also works on your end, I'll resolve.

examples/g4beginner/README.md
jwi3: See above. This should also be fixed now.

examples/g4beginner/CMakeLists.txt
jwi3: Fixed.

examples/g4beginner/g4beginner.cc
jwi3: Should I set `INSTALL_EXAMPLES` in our `CMakeLists.txt` then?

jwi3: I removed the macro invocation and adjusted the Readme instructions so that the visualization is called from the Geant prompt.

examples/g4beginner/scripts/g4cmp_electrons_interface_hits.C
```diff
@@ -0,0 +1,146 @@
...
+  gStyle->SetTitleFillColor(kWhite);
+
+  if (nShots <= 0) {
+    std::cerr << "Error: nShots must be > 0" << std::endl;
+    return;
+  }
+
+  TString yLabel = "Drift electron count / " + primaryType;
+
+  TFile *f = TFile::Open("g4cmp_electrons.root");
```
BirgitZatschler: It would be more flexible to not hardcode the input file name here, but instead pass it as a parameter to the main function. Then, you could also re-use the string for the output png files.

Otherwise it might not be clear to the user which ROOT file is being processed. And it might not be obvious to the user why the file had been copied beforehand `cp g4cmp_electrons.root g4cmp_electrons_100um.root` and later on again `cp g4cmp_electrons.root g4cmp_electrons_300um.root` if these files are not used.

examples/g4beginner/README.md
```diff
@@ -0,0 +1,265 @@
...
+
+In the fourth block, note that phonons are omitted (or killed) in this simulation. This is done simply to speed up the collection of electron statistics. The fifth block controls the mean free paths for electron and hole trapping. We will vary these momentarily to see how they affect the charge dynamics. Finally, we use 100 shots here to obtain results quickly, with the tradeoff of poorer statistics. For the plots shown below, I increased the number of shots to 1000; on my machine, that took about 5 minutes to run.
+
+Once the code has finished running, we will use ROOT to launch a plotting script called `g4cmp_electron_interface_hits.C`,
+
+```bash
+cp g4cmp_electrons.root g4cmp_electrons_100um.root
+root -l 'g4cmp_electrons_interface_hits.C("e-",100)'
```
BirgitZatschler: But this should be now root -l 'g4cmp_electrons_interface_hits.C("e-",1000)' if we followed the advice to increase the number of shots to 1000, no? In the best case the script would read this from the ROOT output instead of having to rely on the user to provide the right value here.
I was impatient and actually only ran with 100 shots because I didn't want to wait 5 minutes.

examples/g4beginner/README.md
```diff
@@ -0,0 +1,265 @@
...
+
+Let's now see how changing the trapping mean free path value affects the results.  In the new macro, `g4cmp_electrons_300um.mac`, the only change is that `eTrappingMFP` is increased from 100 μm to 300 μm. Now run the simulation with the longer trapping mean free path, back up your new output, and run the plotting script again:
+
+```bash
+./g4beginner g4cmp_electrons_300um.mac > g4cmp_electrons_300um.out
+cp g4cmp_electrons.root g4cmp_electrons_300um.root
+root -l 'g4cmp_electrons_interface_hits.C("e-",100)'
```
BirgitZatschler: In the corresponding default macro settings in `g4cmp_electrons_300um.mac`, we run 100 events, which matches with the 100 here. But in your example plots below you ran 1000 events, right? Maybe you want to make this consistent and add a note that you could increase this to 100 again. 100 events do look okay enough for the impatient person or slow machines.

examples/g4beginner/README.md
```diff
@@ -0,0 +1,265 @@
...
+<img src="figures/interface_hits_summary_phonons_100um.png" width="100%">
+
+From this we can compare the behavior of the various flavors of phonons: longitudinal, transverse fast, and transverse slow. These phonon counts reveal that the longitudinal branch is the least populated and the slow transverse branch is the most populated. Phonons are continually down-converting from longitudinal to transverse, so what we see is expected: an inverse proportionality between the population and the phonon energy.
+
+Finally, let us explore the interdependence between the electron and phonon behavior. Let's again change the trapping mean free path to 300 𝜇m, rerun, and regenerate the plots:
+```bash 
+./g4beginner g4cmp_phonons_300um.mac > g4cmp_phonons_300um.out
+root -l 'g4cmp_phonons_interface_hits.C("e-",100)'
+```
+Notably, the phonon distributions are not too different for electron trapping MFP values of 100 vs 300 𝜇m.  Why is that?
```
BirgitZatschler, line 250: I hope you are going to answer this question during the workshop :)

examples/g4beginner/README.md
```diff
@@ -0,0 +1,265 @@
...
+
+The means of these event-level distributions summarize how much pair-breaking-capable phonon population is generated by a single primary radiation event and transported to the Si/Al boundary. Their magnitudes can be much larger than the energy or count associated with any individual tracked phonon because they represent weighted sums over the full phonon cascade produced in one event. This is physically reasonable: a single energetic radiation interaction in the silicon substrate can generate many phonons, and one phonon with energy above 2Δ can in principle contribute to more than one broken Cooper pair through further downconversion and relaxation cascades. Although we are not yet simulating quasiparticles explicitly in the aluminum film, these weighted above-threshold phonon quantities are consistent with the framework of the attached paper. In the language of Vepsäläinen et al., they provide a microscopic proxy for the radiation-induced energy input that would feed quasiparticle generation in the superconductor, ultimately contributing to the excess quasiparticle density $x_{qp}$ and therefore to increased qubit relaxation rate $\Gamma_1$​
```
BirgitZatschler, line 268: I'm missing a hint here to run the script `g4cmp_phonons_pairbreaking_summary.C` to get these plots.

examples/g4beginner/G4Macros/g4cmp_phonons_100um.mac
```diff
@@ -0,0 +1,22 @@
+/run/initialize
+/control/verbose 2
+/run/verbose 2
+/event/verbose 0
+/tracking/verbose 1
```
BirgitZatschler, line 5: I recommend putting the verbosity to 0 here if you don't work with that output anyway. It created ~ GB large output files and slows down the simulation unnecessarily.

Instead, it could be useful to just add `/run/printProgress 1` to get a feeling how long the simulation runs.

jwi3: Fixed, although now it just makes it glaringly obvious that Jesse didn't want to mess with border/skin surfaces. Oh well!

examples/g4beginner/README.md
jwi3: Fixed.

examples/g4beginner/README.md
jwi3: I'm not sure why he chose to do it this way, but I'm not inclined to touch it myself since I would have to replace his plots with new ones.

examples/g4beginner/README.md
jwi3: I don't think he wanted users running with 1000 particles, I think he just did it himself for the runs shown in the plots. I think he was just trying to be transparent there.

examples/g4beginner/scripts/g4cmp_electrons_interface_hits.C
jwi3: Done.

examples/g4beginner/G4Macros/g4cmp_electrons_100um.mac
```diff
@@ -2,7 +2,7 @@
 /control/verbose 2
 /run/verbose 2
 /event/verbose 0
-/tracking/verbose 1
+/run/printProgress 1
```
kelseymh: Don't change it at this point (maybe in the future).  A lot of these "beginner" scripts would have `/tracking/verbose 0` explicitly at the top, so users have something already in place they can fiddle with.  I'd put the `/run/printProgress` (which I always like to have for myself) line right before `beamOn`.

examples/g4beginner/README.md
```diff
@@ -0,0 +1,268 @@
...
+```bash 
+root -l 'g4cmp_phonons_pairbreaking_summary.C("g4cmp_phonons", "e-", 100)'
+```
+
+ <img src="figures/phonons_pairbreaking_summary.png" width="100%">
```
BirgitZatschler: The figure is named "phonon" not "phonons". It doesn't show up in the README.

jwi3: Fixed.

examples/g4beginner/G4Macros/g4cmp_electrons_100um.mac
jwi3: Both commands are now in all macros (except `*vis.mac`).

PR #88, opened 2026-06-12
G4 cmp 628

emichaud4:
Small fix so charge transport can be added for materials with conduction band at Gamma (direct gap). With this change, valleyDir 0 0 0 will addValley Identity matrix instead of the faulty (0,0,0   ,0,1,0,   0,0,1).

### DIFF
--- a/library/src/G4LatticeLogical.cc
+++ b/library/src/G4LatticeLogical.cc
@@ -60,6 +60,7 @@
 // 20240510  E. Michhaud -- Add function to compute L0 from other parameters
 // 20250904  R. Linehan -- Linking Tcrit to Delta0 for superconductors
 // 20250905  G4CMP-500 -- Removing non-fundamental superconductor parameters
+// 20260612  E. Michaud -- If ValleyDir 0 0 0, then AddValley Identity Matrix
 
 #include "G4LatticeLogical.hh"
 #include "G4CMPPhononKinematics.hh"	// **** THIS BREAKS G4 PORTING ****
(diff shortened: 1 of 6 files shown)


kelseymh (approved):
We will not merge this onto develop until after RISQ.

library/src/G4LatticeLogical.cc
```diff
@@ -845,7 +846,13 @@ void G4LatticeLogical::AddValley(const G4ThreeVector& valleyDirVec, G4bool antiv
...
+
+  // If vx=vy=vz=0 use Identity matrix
+  if (vdir.mag()<1e-10) {
+    vdir = G4ThreeVector(1,0,0);
+    a=0; b=1; d=0; e=0; f=1;
+  }
+
   // Store the valley's rotation matrix, its inverse and the valley's direction
   fValley.resize(fValley.size()+1);
   fValley.back().setRows(vdir, G4ThreeVector(a,b,0), G4ThreeVector(d,e,f));
```
kelseymh, line 859: As discussed in the JIRA ticket G4CMP-628, I think it would be best if the Gamma valley did not have a corresponding antivalley.

kelseymh (changes requested):
In addition to (and because of) the change in `G4LatticeLogical`, I would like to propose adding a modification to `G4CMPInterValleyScattering::GetMeanFreePath()`, so that if `NumberOfValleys()==1`, the MFP overrides the base class calculation and returns `DBL_MAX`.   That will prevent IV scattering from being invoked _at all_ if there's only the one Gamma valley in play.

library/src/G4LatticeLogical.cc
emichaud4: Finally, I decided to implement this feature in G4LAtticeLogical::AddValley instead of G4LatticeReader.

kelseymh (approved):

PR #91, opened 2026-06-18
Add sanity check for charge parameters when G4LatticeLogical is created.

kelseymh:
Emile discovered that if we don't explicitly define at least one valley, the `fValley` list is empty, and the result triggers a segmentation fault when new electron tracks are created and can't be assigned a valley index.

For G4CMP-633, I am adding a consistency check as part of `G4LatticeLogical::Initialize()`, to ensure that a minimal set of charge transport parameters are provided.  For the specific issue of `fValley` being empty, I will add a single (0,0,0) valley, corresponding to an isotropic direct bandgap semiconductor.  I am also adding a set of additional checks to let developers know explicitly if their brand new config.txt file is sufficient or broken.

As a side change, I've also done a bit of compressing of Ryan's QP parameter consistency check.

### DIFF
--- a/library/src/G4LatticeLogical.cc
+++ b/library/src/G4LatticeLogical.cc
@@ -81,7 +81,7 @@ G4LatticeLogical::G4LatticeLogical(const G4String& name)
     fPermittivity(1.), fElasticity{}, fElReduced{}, fHasElasticity(false),
     fpPhononKin(0), fpPhononTable(0),
     fA(0), fB(0), fLDOS(0), fSTDOS(0), fFTDOS(0), fTTFrac(0), 
-    fBeta(0), fGamma(0), fLambda(0), fMu(0),
+    fBeta(0), fGamma(0), fLambda(0), fMu(0), fDebye(0),
     fSC_Tau0_qp(DBL_MAX), fSC_Tau0_ph(DBL_MAX),
     fVSound(0.), fVTrans(0.), fL0_e(0.), fL0_h(0.), 
     mElectron(electron_mass_c2/c_squared),
@@ -1123,8 +1123,6 @@ void G4LatticeLogical::DumpList(std::ostream& os,
 // Check for completeness of charge transport parameters
 
 void G4LatticeLogical::CheckLatticeChargeParameters() {
-  G4bool goodChg = true;
-
   // No bandgap will suppress e/h creation, but allow single tracking
   if (fBandGap <= 0. && fPairEnergy <= 0.) {
     if (verboseLevel) {
@@ -1137,44 +1135,49 @@ void G4LatticeLogical::CheckLatticeChargeParameters() {
   if (fPairEnergy < fBandGap) {
     G4ExceptionDescription msg;
     msg << "Lattice " << fName << " has pair energy " << fPairEnergy/eV
-	<< " eV below bandgap " << fBandGap << " eV.";
+	<< " eV, which is below bandgap " << fBandGap/eV << " eV.";
     G4Exception("G4LatticeLogical", "Lattice008", FatalException, msg);
-    goodChg = false;
   }
 
   // If charge masses are not defined, remind user they're free electrons
-  if (goodChg && (fabs(fElectronMass-mElectron)/mElectron < 1e-3 ||
-		  fabs(fHoleMass-mElectron)/mElectron < 1e-3)) {
+  if (fabs(fElectronMass-mElectron)/mElectron < 1e-3 ||
+      fabs(fHoleMass-mElectron)/mElectron < 1e-3) {
     G4ExceptionDescription msg;
     msg << "Lattice " << fName << " does not have charge carrier masses set."
 	<< "\n Bare electron mass assumed.";
     G4Exception("G4LatticeLogical", "Lattice009", JustWarning, msg);
   }
 
+  // Sensible charge transport must include finite scattering lengths
+  if (fL0_e <= 0. || fL0_h <= 0.) {
+    G4ExceptionDescription msg;
+    msg << "Lattice " << fName << " has no charge scattering L0 defined.";
+    G4Exception("G4LatticeLogical", "Lattice010", FatalException, msg);
+  }
+
+  // Speed of sound must be defined, to avoid infinite Luke scattering
+  if (fVSound <= 0. && fVTrans <= 0.) {
+    G4ExceptionDescription msg;
+    msg << "Lattice " << fName << " has no speed of sound defined.";
+    G4Exception("G4LatticeLogical", "Lattice011", FatalException, msg);
+  }
+
   // Without valleys, assume a direct gap semiconductor with Gamma 'valley'
-  if (goodChg && fValley.size() < 1U) {
+  // TODO: Move to CheckIVConsistency() under G4CMP-404
+  if (fValley.size() < 1U) {
     G4cout << "Lattice " << fName << " assuming direct gap semiconductor,"
 	   << " using Gamma valley (0,0,0)." << G4endl;
 
     AddValley(G4ThreeVector(0,0,0));
   }
 
-  // Sensible charge transport must include finite scattering lengths
-  if (goodChg && (fL0_e <= 0. || fL0_h <= 0.)) {
-    G4ExceptionDescription msg;
-    msg << "Lattice " << fName << " has no charge scattering L0 defined.";
-    G4Exception("G4LatticeLogical", "Lattice010", FatalException, msg);
-    goodChg = false;
-  }
-
   // Multiple valleys ought to have IV scattering, but may not
-  if (goodChg && fValley.size() > 1U && fIVModel.empty()) {
+  // TODO: Move to CheckIVConsistency() under G4CMP-404
+  if (fValley.size() > 1U && fIVModel.empty()) {
     G4ExceptionDescription msg;
     msg << "Lattice " << fName << "has no intervalley scattering model.";
-    G4Exception("G4LatticeLogical", "Lattice011", JustWarning, msg);
+    G4Exception("G4LatticeLogical", "Lattice012", JustWarning, msg);
   }
-
-  // TODO: Should we be checking the different IV model parmaeters?
 }
 
 
--- a/library/src/G4LatticeLogical.cc
+++ b/library/src/G4LatticeLogical.cc
@@ -1165,8 +1165,10 @@ void G4LatticeLogical::CheckLatticeChargeParameters() {
   // Without valleys, assume a direct gap semiconductor with Gamma 'valley'
   // TODO: Move to CheckIVConsistency() under G4CMP-404
   if (fValley.size() < 1U) {
-    G4cout << "Lattice " << fName << " assuming direct gap semiconductor,"
-	   << " using Gamma valley (0,0,0)." << G4endl;
+    G4ExceptionDescription msg;
+    msg << "Lattice " << fName << " has no valley definitions.\n"
+	<< " Assuming direct gap semiconductor with Gamma valley.";
+    G4Exception("G4LatticeLogical", "Lattice012", JustWarning, msg);
 
     AddValley(G4ThreeVector(0,0,0));
   }
@@ -1176,7 +1178,7 @@ void G4LatticeLogical::CheckLatticeChargeParameters() {
   if (fValley.size() > 1U && fIVModel.empty()) {
     G4ExceptionDescription msg;
     msg << "Lattice " << fName << "has no intervalley scattering model.";
-    G4Exception("G4LatticeLogical", "Lattice012", JustWarning, msg);
+    G4Exception("G4LatticeLogical", "Lattice013", JustWarning, msg);
   }
 }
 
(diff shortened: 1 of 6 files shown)


library/src/G4LatticeLogical.cc
```diff
@@ -1116,14 +1120,72 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
+}
+
+
 //This checks the results of the created lattice to understand if all relevant
 //parameters have been added. We can put whatever we want here, but for now I
 //want to say that if any of the superconductor lattice parameters have been
 //added, we want to make sure that all of them have, and alert users that if
 //that isn't true, then they should make it true. This forced execution stop
 //occurs here for the "fundamental" parameters and in SCUtils for the "physical-
 //lattice-dependent" ones
```
kelseymh: I should consolidate this narrative,. as well as inserting the required space after the comment markers.

library/src/G4LatticeLogical.cc
```diff
@@ -1116,14 +1120,72 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
-void G4LatticeLogical::CheckLatticeForSCCompleteness() {
+void G4LatticeLogical::CheckLatticeForSCCompleteness() const {
   //Check to see if any of the SC parameters are not at their default values,
   //i.e. if they have been set.
```
kelseymh: Add missing space after comment markers.

library/src/G4LatticeLogical.cc
```diff
@@ -1136,12 +1198,9 @@ void G4LatticeLogical::CheckLatticeForSCCompleteness() {
       //Throw a warning that there are outstanding SC parameters that are not
       //set.
```
kelseymh: Remove this comment.  It's entirely redundant with the contents of the action.

library/src/G4LatticeLogical.cc
```diff
@@ -1116,14 +1120,72 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
+      G4cout << "Lattice " << fName << " does not have bandgap or pair energy"
+	     << " defined.  No charge production." << G4endl;
+    }
+  }
+
+  // If bandgap is defined, pair energy must be larger
+  if (fPairEnergy < fBandGap) {
+    G4ExceptionDescription msg;
+    msg << "Lattice " << fName << " has pair energy " << fPairEnergy/eV
+	<< " eV below bandgap " << fBandGap << " eV.";
```
emichaud4: Should we have fBandGap/eV?

emichaud4: The current wording make it sounds like the difference (fpairEnergy - fBandGap) is ... eV and not that the actual fpairEnergy is ... eV.
 
Maybe add a ; between eV and below bandgap or something like that to emphasize that the fpairEnergy is. ... eV, which is below the fBandGap ... eV ?

library/src/G4LatticeLogical.cc
```diff
@@ -1116,14 +1120,72 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
+	<< "\n Bare electron mass assumed.";
+    G4Exception("G4LatticeLogical", "Lattice009", JustWarning, msg);
+  }
+
+  // Without valleys, assume a direct gap semiconductor with Gamma 'valley'
+  if (goodChg && fValley.size() < 1U) {
+    G4cout << "Lattice " << fName << " assuming direct gap semiconductor,"
+	   << " using Gamma valley (0,0,0)." << G4endl;
+
+    AddValley(G4ThreeVector(0,0,0));
```
emichaud4, line 1170: This will only works once G4CMP-628 is pushed into develop, but I agree, it's the right way to compute the Identity matrix

emichaud4: it's also missing a bool argument about whether or not it's an anti-valley. We should have something like :

AddValley(G4ThreeVector(0,0,0), false);

emichaud4: false meaning it's a valley, true is for anti-valley

library/src/G4LatticeLogical.cc
```diff
@@ -1116,14 +1120,72 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
+  }
+
+  // Sensible charge transport must include finite scattering lengths
+  if (goodChg && (fL0_e <= 0. || fL0_h <= 0.)) {
```
emichaud4: The scattering length is computed from physical constants (acoustic deformation potential, mass, etc.) in `G4LatticeLogical::ComputeL0` and is no longer needed in `configt.txt`. If l0 is not in config.txt or is set to 0, then we compute L0 in G4LatticeManager::LoadLattice (I know it's a terrible place. I moved it to `G4LAtticeLogical::Initialize` in G4MCP-404). I should also change the condition to `if (fL0_e <= 0) : computeL0;` in G4CMP-404. All to say, I would probably remove L0 from the needed charge parameters. Let me know what you think!

library/src/G4LatticeLogical.cc
```diff
@@ -1116,14 +1120,72 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
+    G4ExceptionDescription msg;
+    msg << "Lattice " << fName << " has no charge scattering L0 defined.";
+    G4Exception("G4LatticeLogical", "Lattice010", FatalException, msg);
+    goodChg = false;
+  }
+
+  // Multiple valleys ought to have IV scattering, but may not
+  if (goodChg && fValley.size() > 1U && fIVModel.empty()) {
```
emichaud4: I think the default IV model is Quadratic. I'm curious, what happens if we simply remove the ivModel line in config.txt? I'm simply wondering if Quadratic will be choosen by default and fIVModel won't be empty?

I should add that if no IV Quad (or even Linear) parameters are in config.txt, IV rate will be 0 and no IV scattering will happen.

library/src/G4LatticeLogical.cc
```diff
@@ -1116,14 +1120,72 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
+    G4ExceptionDescription msg;
+    msg << "Lattice " << fName << "has no intervalley scattering model.";
+    G4Exception("G4LatticeLogical", "Lattice011", JustWarning, msg);
+  }
+
+  // TODO: Should we be checking the different IV model parmaeters?
```
emichaud4: Strangely enough, the only IV parameters "needed" to run is ivEnergy because it's used in Emin calculations in `IVRateLinear` and` IVRateQuadratic`. We should probably add a check that if `ivEnergy` is not `config.txt`, then set it to 0?

library/src/G4LatticeLogical.cc
emichaud4: When G4CMP-628 will be pushed to develop, direct band gap materials with only valleyDir 0 0 0 in config.txt will have two Identity matrices, meaning the fValley.size() > 1U condition will be true. I suggest we keep this line as it is and I will fix G4CMP-628 so only one valley matrix is stored in fValley.

library/src/G4LatticeLogical.cc
emichaud4: I would add the three following checks for IV parameters depending on which IV model got selected (the name of first principles might be IVRate in develop branch but I'm not sure) : 
```
// Intervalley phonon parameters missing (from Jacoboni matrix elements)
if (goodChg && fValley.size() > 1U && !fIVModel.empty() && fIVModel == "Matrix" &&
    ( fIVDeform.empty() || fIVEnergy.empty()) { ... }

// Linear model selected but rate parameters missing
if (goodChg && fValley.size() > 1U && !fIVModel.empty() &&
    fIVModel == "Linear" &&
    (fIVLinRate0 == 0. || fIVLinRate1 == 0. || fIVLinPower == 0.)) { ... }

// Quadratic model selected but rate parameters missing
if (goodChg && fValley.size() > 1U && !fIVModel.empty() &&
    fIVModel == "Quadratic" &&
    (fIVQuadRate == 0. || fIVQuadField == 0. || fIVQuadPower == 0.)) { ... }
```

emichaud4: with the check IV consistency in G4CMP-404, we only need to check if fIVDeform and fIVenergy are empty. The new function CheckIVconstistency will look if those are the same size. I will also add the other iv parameters from first principles in the check ( fIVphononmode.empty(), ...) in G4CMP-404

library/src/G4LatticeLogical.cc
```diff
@@ -1116,14 +1120,72 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
+}
+
```
emichaud4, line 1181: I would probably add a check on speed of sound and deformation potentials? if both are not specified, they are set to 0. This results in either luke scattering rate =0 (Dac =0) or infinity (vsound=0)

library/src/G4LatticeLogical.cc
kelseymh: It looks like the argument is optional; see `G4LatticeLogical.hh`:
```
  void AddValley(const G4ThreeVector&, G4bool antival=false);
```

emichaud4: Nvm then!

library/src/G4LatticeLogical.cc
kelseymh: The `Initialize()` testing is not about the contents of the `config.txt` file.  It is about the _data members_ of the `G4LatticleLogical` class.  The test here involving the two L0 _values_ is valid whether they were set explictly, or if they were computed previously from `acDeform`.

library/src/G4LatticeLogical.cc
kelseymh: For the "default IV model", you're correct.  In the `G4LatticeLogical` constructor, we have the initializer
```
fIVModel(G4CMPConfigManager::GetIVRateModel())
```
So it defaults to whatever is set in the ConfigManager.  Right now that's Quadratic, but it could change (which is why it's _not_ hardcoded here).  So in practice, the test condition is unnecessary, because of other features of the code.  But it's still a valid state check, which is what this function is about.

kelseymh: > When G4CMP-628 will be pushed to develop, direct band gap materials with only valleyDir 0 0 0 in config.txt will have two Identity matrices, meaning the fValley.size() > 1U condition will be true.

I agree with you that this should be fixed in G4CMP-628.  I will make a comment in that JIRA ticket.

library/src/G4LatticeLogical.cc
emichaud4: We should keep the check then!

library/src/G4LatticeLogical.cc
emichaud4: Thank you Mike, that answers my questions. No action needed for this part

library/src/G4LatticeLogical.cc
kelseymh: > The new function CheckIVconstistency will look if those are the same size. I will also add the other iv parameters from first principles in the check ( fIVphononmode.empty(), ...) in G4CMP-404

Having a separate function for the IV parameters (for example, in different model combinations) is a great idea.   The IV-specific checks here should move into that function.  Please make a new, separate _sub-task_ under G4CMP-404 for this.  Specify (via "Link") that it requires G4CMP-633 to be completed first.

I will not add any more IV checks here, and I will modify the "TODO" comment to indicate that it will be done as part of the G4CMP-404 project.

library/src/G4LatticeLogical.cc
kelseymh: See my followup to a previous comment: we shouldn't check explicitly for the deformation potentials: the L0 data members may be computed from them, and that is what we test here (after the computation has been done).

Testing for the two speeds of sound makes sense.  In principle, those ought to be computable from the underlying crystal information (stress-strain tensor, material density, etc.), but whether computed or preset, this function should test whether the data members are sensible.  Since vsound == 0 would result in divide by zero errors at runtime, it will trigger a fatal exception.

library/src/G4LatticeLogical.cc
kelseymh: > the only IV parameters "needed" to run is ivEnergy because it's used in Emin calculations in IVRateLinear and IVRateQuadratic. We should probably add a check that if ivEnergy is not config.txt, then set it to 0?

That's relatively new, I think (... G4CMP-358), and is why the zero-volt rates don't explode any more!  Either we should require the thresholds to be filled, or the Threshold() functions should include an `.empty()` check, to return zero threshold.  I have created G4CMP-636 for this.

library/src/G4LatticeLogical.cc
```diff
@@ -1116,32 +1120,77 @@ void G4LatticeLogical::DumpList(std::ostream& os,
...
-	  << "Please fill in Tau0_qp and Tau0_ph for all superconductors you "
-	  << "want to use before continuing.";
-      G4Exception("G4LatticeLogical::CheckLatticeForSCCompleteness", "Lattice007",
-		  FatalException, msg);
+// Check for completeness of charge transport parameters
+
+void G4LatticeLogical::CheckLatticeChargeParameters() {
+  // No bandgap will suppress e/h creation, but allow single tracking
+  if (fBandGap <= 0. && fPairEnergy <= 0.) {
+    if (verboseLevel) {
```
kelseymh, line 1125: Question for @emichaud4 :  Should I keep the verbosity check, or have it print the warning no matter what?  It means users will start getting this warning for sapphire and the other new materials, which could get annoying.

emichaud4: I would keep the verbosity. I think a lot of people are running new materials only for phonon transport and that would be annoying for them hahaha!

PR #92, opened 2026-06-29
Add CheckIV consistency (missing or invalid parameters) to G4CMP-404

emichaud4:


emichaud4:
There should be more differences between the two, especially in G4LAtticeLogical::CheckIVConsistency. Not sure why they are not showing

emichaud4:
I was thinking I could do the following to properly show the differences between 404 and 638 since 638 was branched off 404 but 404 was modified afterwards :

```
git checkout G4CMP-638
git rebase origin/G4CMP-404
git push --force-with-lease origin G4CMP-638
```

kelseymh:
Did you merge *develop* branch onto G4CMP-404 after *g4cmp-V10-04-01* was released? ... It looks like you merged develop just before I did the V10-04-01 release, which should be just fine.  Please don't rebase.  You already pushed your version of G4CMP-638 upstream; rebasing now would lead to conflicts.

library/src/G4LatticeLogical.cc
```diff
@@ -66,6 +66,7 @@
 // 20260617  G4CMP-633 -- Cross-check that sufficient charge parameters are set
 // 20260618  G4CMP-637 -- Move calculation of L0 from acDeform out of Manager.
 // 20260618  E. Michaud -- If ValleyDir 0 0 0, then AddValley Identity Matrix
+// 20260629  G4CMP-638 -- Add CheckIV consistency : missing or invalid parameters.
```
kelseymh: Maybe say "Update CheckIVConsistency()"?  This branch shouldn't be creating it, just adding some more error cases.

library/include/G4LatticeLogical.hh
```diff
@@ -50,6 +50,7 @@
 
 // 20260617  G4CMP-633 -- Cross-check that sufficient charge parameters are set
 // 20260618  G4CMP-637 -- Move calculation of L0 from acDeform out of Manager.
+// 20260629  G4CMP-638 -- Add CheckIV consistency : missing or invalid parameters.
```
kelseymh: Maybe say "Update CheckIVConsistency()"? This branch shouldn't be creating it, just adding some more error cases.

library/src/G4LatticeLogical.cc
kelseymh, line 275: I'd suggest moving these lines above the "CheckXYZ()" calls.  That way, all the "added calculations" are together, and are done before any consistency checks which need those calculations.

kelseymh: Can you remove this extra blank line?

kelseymh: bandP doesn't need to be in parentheses.

library/src/G4LatticeLogical.cc
```diff
@@ -999,20 +1000,91 @@ SetIVDeform(ivdeformtest);
...
+  }
+
+  // Quadratic model selected but rate parameters missing
+  if (fValley.size() > 1U && fIVModel == "Quadratic" &&
+    (fIVQuadRate == 0. || fIVQuadField == 0. || fIVQuadExponent == 0.)) {
+    G4ExceptionDescription msg; msg << "Lattice " << fName 
+    << " has missing intervalley scattering (Quadratic) parameters.";
+    G4Exception("G4LatticeLogical", "Lattice020", JustWarning, msg);
+  }
+
```
kelseymh, line 1079: Where are the new consistency checks posted in the G4CMP-638 description?

kelseymh (changes requested):
I added some line-specific comments, but it appears as though the actual new code from G4CMP-638 isn't here yet.  Maybe you didn't push?

library/src/G4LatticeLogical.cc
emichaud4: Done!

emichaud4: Done!

emichaud4: Removed the parentheses!

library/src/G4LatticeLogical.cc
emichaud4: I tried to moved the code a bit so that it would show up in the PR. My mistake, I pretty much "finished" 638 while I was still in branch 404 and realized my mistake too late. To not loose anything, I pushed 404 and then branched 638 off 404 and "undid" the changes in 404 but the changes only appear if I compare 404 to 638 and not the other way around because of that :(

library/src/G4LatticeLogical.cc
```diff
@@ -1011,79 +1010,75 @@ void G4LatticeLogical::CheckIVConsistency() {
...
-    G4ExceptionDescription msg;
-    msg << "Lattice " << fName << " has no intervalley scattering model.";
-    G4Exception("G4LatticeLogical", "Lattice013", JustWarning, msg);
-  }
-    
-  // Intervalley phonon parameters (matrix) missing or not valid
-  if (fValley.size() > 1U && fIVModel == "Matrix") {
-    if (fIVPhononMode.empty() || fIVDeform.empty() || fIVEnergy.empty() ||
-      fIVNValleys.empty() || fIVOrder.empty() || fIVFGScattering.empty()) {
+  if (fValley.size() > 1U) {
```
kelseymh: Just a suggestion, to simplify reading the code for humans.  Change this line to:
```
  if (fValley.size() < 2U) return;    // No IV scattering possible
```
and move everything below it out one scoping level.

library/src/G4LatticeLogical.cc
```diff
@@ -1011,79 +1010,75 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+    // Multiple valleys ought to have IV scattering, but may not
+    if (fIVModel.empty()) {
       G4ExceptionDescription msg; msg << "Lattice " << fName
-      << " has missing intervalley scattering (Matrix) parameters.";
+      << " has no intervalley scattering model.";
```
kelseymh: Maybe say "no preset intervalley scattering model."?

library/src/G4LatticeLogical.cc
```diff
@@ -1011,79 +1010,75 @@ void G4LatticeLogical::CheckIVConsistency() {
...
-      }
-
-      static const std::set<G4String> PhononValidModes = {"TA", "LA", "TO", "LO"};
-      for (size_t i = 0; i < n; ++i) {
-        if (PhononValidModes.find(fIVPhononMode[i]) == PhononValidModes.end()) {
-          G4ExceptionDescription msg; msg << "Lattice " << fName
-          << " has unknown phonon mode.";
-          G4Exception("G4LatticeLogical", "Lattice016", JustWarning, msg);
+    }
+    // Quadratic model selected but rate parameters missing
```
kelseymh: Add a blank line above this.

library/src/G4LatticeLogical.cc
```diff
@@ -1011,79 +1010,75 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+    if (fIVModel == "Quadratic" &&
```
kelseymh: I think you want to do the parameter check regardless of the default value.  Users can _choose_ "Quadratic" or "Linear" or "Matrix" via macro command.  You want to test that the parameters for a given model are either all there and sensible, or all not there.

library/src/G4LatticeLogical.cc
```diff
@@ -1011,79 +1010,75 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+      (fIVQuadRate == 0. || fIVQuadField == 0. || fIVQuadExponent == 0.)) {
+      G4ExceptionDescription msg; msg << "Lattice " << fName
+      << " has missing intervalley scattering (Quadratic) parameters.";
+      G4Exception("G4LatticeLogical", "Lattice015", JustWarning, msg);
+    }
+    // Intervalley phonon parameters (matrix) missing or not valid
+    if (fIVModel == "Matrix") {
```
kelseymh: As noted above, you want to test that the set of Matrix parameters are consistent with each other, regardless of whether it's the default or not.

library/src/G4LatticeLogical.cc
```diff
@@ -1011,79 +1010,75 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+      if (fIVPhononMode.empty() ||  fIVOrder.empty() || fIVEnergy.empty() ||
+        fIVNValleys.empty() || fIVDeform.empty() || fIVFGScattering.empty()) {
+        G4ExceptionDescription msg; msg << "Lattice " << fName
+        << " has missing intervalley scattering (Matrix) parameters.";
+        G4Exception("G4LatticeLogical", "Lattice016", JustWarning, msg);
+      } else {
+        const size_t n = fIVDeform.size();
+          G4cout << n << G4endl;
```
kelseymh: Delete this line (or hide it behind verbosity, but I'd prefer just removing it.

library/src/G4LatticeLogical.cc
```diff
@@ -1011,79 +1010,75 @@ void G4LatticeLogical::CheckIVConsistency() {
...
 
-      static const std::set<G4String> IVFGmode = {"f", "g"};
-      for (size_t i = 0; i < n; ++i) {
-        if (IVFGmode.find(fIVFGScattering[i]) == IVFGmode.end()) {
-          G4ExceptionDescription msg; msg << "Lattice " << fName
-          << " has unknown (f or g) IV scattering type.";
-          G4Exception("G4LatticeLogical", "Lattice018", JustWarning, msg);
+        static const std::set<G4String> PhononValidModes = {"TA", "LA", "TO", "LO"};
+        for (size_t i = 0; i < n; ++i) {
+          if (PhononValidModes.find(fIVPhononMode[i]) == PhononValidModes.end()) {
```
kelseymh: There should be a way to do this without the loop, using an STL algorithm.  Something similar to the std::string function "find_first_not_of".

library/src/G4LatticeLogical.cc
```diff
@@ -1011,79 +1010,75 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+            G4ExceptionDescription msg; msg << "Lattice " << fName
+            << " has unknown phonon mode.";
+            G4Exception("G4LatticeLogical", "Lattice019", JustWarning, msg);
+          }
+        }
+        static const std::set<G4String> IVFGmode = {"f", "g"};
+        for (size_t i = 0; i < n; ++i) {
+          if (IVFGmode.find(fIVFGScattering[i]) == IVFGmode.end()) {
```
kelseymh: See above.

kelseymh (changes requested):

library/src/G4LatticeLogical.cc
emichaud4: I think there is no IV linear parameters for Ge. Which means, the warning message will appear everytime someone runs Ge, even if they selected the quadratic model.

library/src/G4LatticeLogical.cc
emichaud4: Done!

library/src/G4LatticeLogical.cc
emichaud4: done!

library/src/G4LatticeLogical.cc
emichaud4: done!

library/src/G4LatticeLogical.cc
emichaud4: deleted! Sorry, I thought I removed it before pushing when I finished testing my changes with some prints

library/src/G4LatticeLogical.cc
emichaud4: I see what you mean now! I tried something to test for all or none parameters (if all parameters are there : good or if none : good too but if part of the parameters are there : warning ). I removed the check on the IV model. 

Only "problem" I see is if Matrix model is selected (via config.txt or macro command) and one of the vector of parameters is missing, the simulation crashes instantly after the warning message.

library/src/G4LatticeLogical.cc
emichaud4: see my response above!

library/src/G4LatticeLogical.cc
emichaud4: I now use std::any_of. Let me know it this works!

library/src/G4LatticeLogical.cc
emichaud4: I now use std::any_of. Let me know it this works!

library/src/G4LatticeLogical.cc
kelseymh: In that case, you should treat that as FatalException, not as warning.  This should be the case for all three -- if any of the parameters are set, but not all of them, that should kill the job.

library/src/G4LatticeLogical.cc
kelseymh: Did it work for you when you tested it?

library/src/G4LatticeLogical.cc
```diff
@@ -1004,79 +1004,88 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+  // Multiple valleys ought to have IV scattering, but may not
+  if (fIVModel.empty()) {
+    G4ExceptionDescription msg; msg << "Lattice " << fName
+    << " has no preset intervalley scattering model.";
+    G4Exception("G4LatticeLogical", "Lattice013", JustWarning, msg);
+  }
+
+  // Linear model: warn only if partially configured
+  const int nLinear =
+    (fIVLinRate0 != 0.) + (fIVLinRate1 != 0.) + (fIVLinExponent != 0.);
```
kelseymh, line 1024: Clever solution!

library/src/G4LatticeLogical.cc
```diff
@@ -1004,79 +1004,88 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+  } else if (nMatrix == 6) {
+
+    // All six parameters present : make sure they are the same size
+    const size_t n = fIVDeform.size();
+    if (fIVEnergy.size() != n || fIVOrder.size() != n ||
+      fIVNValleys.size() != n || fIVPhononMode.size() != n ||
+      fIVFGScattering.size() != n) {
+      G4ExceptionDescription msg;
+      msg << "Lattice " << fName
+      << " has IV parameters vectors with inconsistent sizes.";
```
kelseymh: Maybe write " has IV parameter vectors (Matrix) with inconsistent sizes."?

library/src/G4LatticeLogical.cc
```diff
@@ -1004,79 +1004,88 @@ void G4LatticeLogical::CheckIVConsistency() {
...
-    if (fIVModel == "Linear" &&
-      (fIVLinRate0 == 0. || fIVLinRate1 == 0. || fIVLinExponent == 0.)) {
+
+    // Make sure IV process order is 0 or 1
+    if (std::any_of(fIVOrder.begin(), fIVOrder.end(),
+      [](int order) { return order != 0 && order != 1; })) {
       G4ExceptionDescription msg; msg << "Lattice " << fName
-      << " has missing intervalley scattering (Linear) parameters.";
-      G4Exception("G4LatticeLogical", "Lattice014", JustWarning, msg);
+      << " has invalid IV order scattering (0 or 1).";
```
kelseymh: I'd write "not 0 or 1" here (that's what's invalid, that it _isn't_ 0 or 1).

library/src/G4LatticeLogical.cc
```diff
@@ -1004,79 +1004,88 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+    // Make sure IV phonon modes are valid (TA, LA, TO or LO)
+    static const std::set<G4String> PhononValidModes = {"TA","LA","TO","LO"};
+    if (std::any_of(fIVPhononMode.begin(), fIVPhononMode.end(),
+      [&](const G4String& mode) {
+        return PhononValidModes.find(mode) == PhononValidModes.end();
+      })) {
       G4ExceptionDescription msg; msg << "Lattice " << fName
-      << " has missing intervalley scattering (Quadratic) parameters.";
-      G4Exception("G4LatticeLogical", "Lattice015", JustWarning, msg);
+      << " has unknown phonon mode.";
```
kelseymh: I'd write, "has unknown IV scattering (Matrix) phonon channel."

library/src/G4LatticeLogical.cc
```diff
@@ -1004,79 +1004,88 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+  }
+
+  // Matrix model: warn only if partially configured
+  const int nMatrix =
+    (!fIVPhononMode.empty()) + (!fIVOrder.empty()) + (!fIVEnergy.empty())
+    + (!fIVNValleys.empty()) + (!fIVDeform.empty()) + (!fIVFGScattering.empty());
+  if (nMatrix > 0 && nMatrix < 6) {
+    G4ExceptionDescription msg; msg << "Lattice " << fName
+    << " has incomplete intervalley scattering (Matrix) parameters.";
+    G4Exception("G4LatticeLogical", "Lattice016", JustWarning, msg);
```
kelseymh: This should probably be FatalException.

library/src/G4LatticeLogical.cc
```diff
@@ -1004,79 +1004,88 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+    G4Exception("G4LatticeLogical", "Lattice014", JustWarning, msg);
+  }
+
+  // Quadratic model: warn only if partially configured
+  const int nQuad =
+    (fIVQuadRate != 0.) + (fIVQuadField != 0.) + (fIVQuadExponent != 0.);
+  if (nQuad > 0 && nQuad < 3) {
+    G4ExceptionDescription msg; msg << "Lattice " << fName
+    << " has incomplete intervalley scattering (Quadratic) parameters.";
+    G4Exception("G4LatticeLogical", "Lattice015", JustWarning, msg);
```
kelseymh: This should probably be FatalException.

library/src/G4LatticeLogical.cc
```diff
@@ -1004,79 +1004,88 @@ void G4LatticeLogical::CheckIVConsistency() {
...
+    G4Exception("G4LatticeLogical", "Lattice013", JustWarning, msg);
+  }
+
+  // Linear model: warn only if partially configured
+  const int nLinear =
+    (fIVLinRate0 != 0.) + (fIVLinRate1 != 0.) + (fIVLinExponent != 0.);
+  if (nLinear > 0 && nLinear < 3) {
+    G4ExceptionDescription msg; msg << "Lattice " << fName
+    << " has incomplete intervalley scattering (Linear) parameters.";
+    G4Exception("G4LatticeLogical", "Lattice014", JustWarning, msg);
```
kelseymh: This should probably be FatalException.

kelseymh (changes requested):

library/src/G4LatticeLogical.cc
emichaud4: Yes, it does. I meant works in the way : "is it okay for you? Is this what you had in mind" Sorry for the confusion. Works was probably not the right choice of word hahaha

library/src/G4LatticeLogical.cc
emichaud4: Done!

library/src/G4LatticeLogical.cc
emichaud4: Done!

library/src/G4LatticeLogical.cc
emichaud4: Done!

library/src/G4LatticeLogical.cc
emichaud4: Done!

library/src/G4LatticeLogical.cc
emichaud4: done!

library/src/G4LatticeLogical.cc
emichaud4: done!

library/src/G4LatticeLogical.cc
emichaud4: done!

kelseymh (approved):

PR #94, opened 2026-07-31
G4CMP-647: Base class for quasiparticle-phonon interactions (KaplanQP)

jwi3:
Adding a base class under G4CMPKaplanQP (G4CMPVKaplanQP) to allow for user-configurable phonon-qp interaction physics. JIRA ticket: https://jira.slac.stanford.edu/browse/G4CMP-647


library/include/G4CMPKaplanQP.hh
```diff
@@ -53,17 +53,18 @@
 //		new DoDirectAbsorption() boolean test.
 // 20240502  G4CMP-344: Reusable vector buffers to avoid memory churn.
 // 20240502  G4CMP-379: Add Fermi-Dirac thermal probability for QP energies.
+// 20250101  G4CMP-439: Create separate debugging file per worker thread;
+//    add EventID and TrackID columns to debugging output.
```
kelseymh, line 57: I'm a bit confused why these are showing up as "new lines."  I put that change in a year and a half ago, and it's been on develop and master since V09-04-00.

library/include/G4CMPKaplanQP.hh
```diff
@@ -76,124 +77,48 @@ namespace G4CMP {
 			  std::vector<G4double>& reflectedEnergies);
 }
 
-class G4CMPKaplanQP {
+class G4CMPKaplanQP : public G4CMPVKaplanQP {
 public:
   G4CMPKaplanQP(G4MaterialPropertiesTable* prop, G4int vb=0);
-  virtual ~G4CMPKaplanQP();
```
kelseymh, line 82: This line shouldn't have been deleted.  Each subclass should (must?) have a virtual dtor (you can't rely on the compiler default, as it's typically not virtual).  If the subclass doesn't have any deletions to do, it should be a no-op:
```
    virtual ~G4CMPKaplanQP() {;}
```

library/include/G4CMPVKaplanQP.hh
```diff
@@ -0,0 +1,166 @@
...
+#include "G4Types.hh"
+#include <fstream>
+#include <vector>
+
+class G4MaterialPropertiesTable;
+
+class G4CMPVKaplanQP {
+public:
+  G4CMPVKaplanQP(G4MaterialPropertiesTable* prop, G4int vb=0);
+  ~G4CMPVKaplanQP();
```
kelseymh: Please make this virtual.  We don't know whether future subclasses will contain objects that need explicit deletion.

library/include/G4CMPVKaplanQP.hh
```diff
@@ -0,0 +1,166 @@
...
+
+  // Turn on diagnostic messages
+  void SetVerboseLevel(G4int vb) { verboseLevel = vb; }
```
kelseymh: I might ask you to make this virtual.  Future subclasses might own objects, or have decorators, where they want to pass verbosity through.  Having the Set function virtual makes that easier.  However, it's not critical, and can always be retrofitted if a real need comes up.

library/include/G4CMPVKaplanQP.hh
```diff
@@ -0,0 +1,166 @@
...
+  void SetVSound(G4double value)              { vSound = value; }
+
+  // Do absorption on sensor/metalization film
+  // Returns absorbed energy, fills list of re-emitted phonons
+  virtual G4double AbsorbPhonon(G4double energy,
+      std::vector<G4double>& reflectedEnergies) const = 0;
+
+protected:
+  // Check that the five required parameters are set to meaningful values
+  G4bool ParamsReady() const {
```
kelseymh: This function should be virtual.  That way, future subclasses can extend it to verify their own additional parameters, calling back to base.

library/src/G4CMPPhononElectrode.cc
```diff
@@ -66,7 +68,7 @@
 // Constructor and destructor
 
 G4CMPPhononElectrode::G4CMPPhononElectrode()
-  : G4CMPVElectrodePattern(), kaplanQP(0) {;}
+  : G4CMPVElectrodePattern(), kaplanQP(0), absorberSet(false) {;}
```
kelseymh: Why do you need the extra flag?  Isn't testing for a null {{kaplanQP}} pointer sufficient?

library/include/G4CMPPhononElectrode.hh
```diff
@@ -67,8 +70,17 @@ public:
...
+  // G4CMPKaplanQP default
+  G4CMPVKaplanQP* getKaplanQP(G4String identifier = "") const;
+  // define phonon-qp interaction mechanism directly
+  // NOTE: PhononElectrode object will take ownership of absorber object
+  // and eventually delete it
+  virtual void RegisterAbsorber(G4CMPVKaplanQP* absorber) const { setKaplanQP(absorber); }
+  // use getKaplanQP to register absorber
+  virtual void RegisterAbsorber(const G4String& identifier) const  { RegisterAbsorber(getKaplanQP(identifier)); }
 protected:
+  virtual void setKaplanQP(G4CMPVKaplanQP* val) const;
```
kelseymh, line 90: Could you add line breaks before each function's prefacing comment?

library/include/G4CMPPhononElectrode.hh
```diff
@@ -78,8 +90,10 @@ protected:
 			 G4ParticleChange& particleChange) const;
 
   // NOTE: "Mutable" because AbsorbAtElectrode() function is const
-  mutable G4CMPKaplanQP* kaplanQP;	// Create instance of QET simulator
+  mutable G4CMPVKaplanQP* kaplanQP;	// Create instance of QET simulator
   mutable std::vector<G4double> phononEnergies;		// Reusable buffer
+  mutable G4bool absorberSet; // so we can tell if user has set kaplanQP to
+  // nullptr intentionally
```
kelseymh: I'm a bit confused by this.  Why isn't the null pointer check sufficient in general?

library/src/G4CMPPhononElectrode.cc
```diff
@@ -93,14 +95,20 @@ AbsorbAtElectrode(const G4Track& track, const G4Step& step,
   }
 
   // Create KaplanQP simulator if not already available
-  if (!kaplanQP) {
-    // Pass temperture through to KaplanQP if no already included
+  if (!kaplanQP && !absorberSet) {
```
kelseymh: It might be worth rethinking this.  I'll start a discussion in the JIRA ticket.

library/src/G4CMPPhononElectrode.cc
```diff
@@ -226,3 +234,26 @@ ProcessReflection(const G4Track& track, const G4Step& step,
...
+void G4CMPPhononElectrode::setKaplanQP(G4CMPVKaplanQP* val) const
+{
+  if (kaplanQP)
+  {
+    delete kaplanQP;
+    kaplanQP = 0;
+  }
+  kaplanQP = val;
+  absorberSet = true;
+}
```
kelseymh: Could you move the open braces to the end of each preceding line?

library/src/G4CMPPhononElectrode.cc
```diff
@@ -226,3 +234,26 @@ ProcessReflection(const G4Track& track, const G4Step& step,
...
+
+// define phonon-qp interaction mechanism via string identifier;
+// G4CMPKaplanQP default
+G4CMPVKaplanQP* G4CMPPhononElectrode::getKaplanQP(G4String identifier) const
+{
+  G4CMPVKaplanQP* val = nullptr;
+  if (identifier.empty() || identifier.find("aplan") != G4String::npos)
+  {
```
kelseymh: Could you move the open braces to the end of each preceding line?

library/src/G4CMPPhononElectrode.cc
```diff
@@ -226,3 +234,26 @@ ProcessReflection(const G4Track& track, const G4Step& step,
...
+    delete kaplanQP;
+    kaplanQP = 0;
+  }
+  kaplanQP = val;
+  absorberSet = true;
+}
+
+// define phonon-qp interaction mechanism via string identifier;
+// G4CMPKaplanQP default
+G4CMPVKaplanQP* G4CMPPhononElectrode::getKaplanQP(G4String identifier) const
```
kelseymh: Unless you plan to do a .to_lower() to be case insensitive, this should be a const reference.

library/src/G4CMPVKaplanQP.cc
```diff
@@ -0,0 +1,228 @@
...
+  //           /
+  //           sqrt((E'*E' - gapEnergy*gapEnergy) *
+  //                ((Energy - E')*(Energy - E') - gapEnergy*gapEnergy));
+  // The shape of the PDF is like a U, so the max values are at the endpoints:
+  // E' = gapEnergy and E' = Energy - gapEnergy
+
+  // Add buffer so first/last bins don't give zero denominator in pdfSum
+  const G4double BUFF = 1000.;
+  G4double xmin = gapEnergy + (Energy-2.*gapEnergy)/BUFF;
+  G4double xmax = gapEnergy + (Energy-2.*gapEnergy)*(BUFF-1.)/BUFF;
```
kelseymh, line 173: I noticed elsewhere you're allowing the possibility of Delta (gapEnergy) = 0.  I'm not sure that's total safe, especially here and in the other PDFs.

library/src/G4CMPVKaplanQP.cc
```diff
@@ -0,0 +1,228 @@
...
+  const G4double gapsq = gapEnergy*gapEnergy;
+  const G4double occupy = 1. - ThermalPDF(E) - ThermalPDF(E-x);
+
+  return ( occupy * (x*(E-x)+gapsq) / sqrt((x*x-gapsq) * ((E-x)*(E-x)-gapsq)) );
+}
+
+G4double G4CMPVKaplanQP::ThermalPDF(G4double E) const {
+  const G4double kT = k_Boltzmann*temperature;
+  return ( (temperature > 0.) ? 1./(exp(E/kT)+1.) : 0. );
+}
```
kelseymh: Since we now have the thermal PDF in `G4CMPUtils` (see `MaxwellBoltzmannPDF`, `ChooseThermalEnergy`, etc.), maybe we should drop this explicit function?

library/src/G4CMPVKaplanQP.cc
```diff
@@ -0,0 +1,228 @@
...
+  // PDF is not integrable, so we can't do an inverse transform sampling.
+  // Instead, we'll do a rejection method.
+  //
+  // PDF(E') = ((Energy-E')*(Energy-E') * (E'-gapEnergy*gapEnergy/Energy))
+  //           /
+  //           sqrt((E'*E' - gapEnergy*gapEnergy);
+
+  // Add buffer so first bin doesn't give zero denominator in pdfSum
+  const G4double BUFF = 1000.;
+  G4double xmin = gapEnergy + gapEnergy/BUFF;
```
kelseymh, line 209: Another case where we either need gapEnergy > 0. protection, or we should not allow it in the first place.

ChangeHistory
```diff
@@ -6,8 +6,8 @@ Releases are tagged on the 'master' branch as "g4cmp-Vxx-yy-zz".
 Features and bug fixes are tagged on the 'develop' branch using the JIRA
 ticket identifier, "G4CMP-nnn", or using the GitHub pull request, "PR-nn".
 
-2026-07-27  G4CMP-650 : Add missing file to library/CMakeLists.txt header list
 2026-07-21  G4CMP-649 : Protect against missing touchable in SurfaceClearance.
+2026-07-10  G4CMP-647 : Institute base class for phonon-qp interactions.
```
kelseymh: Could you move this up a level, with a new date?  It should end up above G4CMP-650 (not sure why I'm not seeing the latest develop here).

kelseymh (changes requested):

library/include/G4CMPVKaplanQP.hh
jwi3: I made the function virtual and enforced gapEnergy > 0 at this stage. Maybe there's a better place to check that value, but this seemed natural.

library/include/G4CMPKaplanQP.hh
kelseymh: @jwi3 Did you cherry-pick files from develop or something?  I'd like to get this inconsistency resolved (or maybe you branched from an older develop, not the HEAD)?

kelseymh (changes requested):

library/include/G4CMPKaplanQP.hh
jwi3: I just re-merged with develop and supposedly fixed the merge conflicts. If the issue is still there, maybe I just need to hand-copy the contents of ChangeHistory and then make my changes.

kelseymh: Okay, then I won't call it a conflict, and I'll call this resolved.

kelseymh (approved):
I think this is ready to go.

Issue #89, opened 2026-06-15
G4CMP overrides -std=c++11


bloer:
In cmake/G4CMPConfig.cmake, there is an explicit hardcoded `set(G4CMP_CXX_FLAGS "-std=c++11")`, which breaks builds with geant4 11. Is this intentional?  I notice in all of the example CMakeLists.txt, there is a block to override CMAKE_CXX_STANDARD, which I think would not be necessary if this hardcoded variable went away.

szatschler:
It looks like the content of the "cmake" folder has last been touched 4 years ago. The standard install via cmake should be governed by the CMakeLists.txt in the main directory which inherits the CXX standard from the present Geant4 installation. The same logic is going to be propagated into the supported G4CMP examples eventually.

However, one concrete action to be taken is to update the existing files under the cmake folder to be useful examples (and clarify their use case better in the README). To my understanding, they are not relevant for the installation of G4CMP itself via cmake (anymore).

If we want to take any action, I'm happy to transcribe this issue into a Jira ticket.

bloer:
Those files are installed by the CMakeLists.txt https://github.com/G4CMP/G4CMP/blob/master/CMakeLists.txt#L100 

If I comment out the line explicitly setting c++11, the only thing needed to compile against g4cmp in an application CMakeLists.txt is 
```
find_package(G4CMP PATHS $ENV{G4CMPINSTALL}/../..)
 include(${G4CMP_USE_FILE})
```

I'm about 90% sure that all the extra checking for `Geant4_CXX_STANDARD` in the examples' CMakeLists.txt are workarounds trying to fix this bug.

szatschler:
That's a good point. I have created a Jira ticket (G4CMP-642) and will check how things can be made more uniform.

szatschler:
This has been addressed on branch G4CMP-642.

szatschler:
Merged onto branch **develop**, tagged **G4CMP-642**, and released on **master** with **G4CMP-V10-05-00**.

PR #98, opened 2026-08-16
G4CMP-642: Derive c++ version dynamically in G4CMP examples.

szatschler:
Addressing Ben's comment from GitHub issue #89. Changes include:
- Turned cmake/UseG4CMP.cmake into a template storage for the variables G4CMP_CXX_STANDARD and G4CMP_VERSION
- Populate G4CMP_CXX_STANDARD and G4CMP_VERSION in the main project's CMakeLists.txt
- Use configure_file() to copy the UseG4CMP.cmake template into the current build directory.
- Copy local UseG4CMP.cmake to the install target
- Read cmake variables from UseG4CMP.cmake in example

Updated all examples (phonon, caustics, charge, sensors, quasiparticles, g4beginner) appropriately.


CMakeLists.txt
```diff
@@ -88,18 +89,35 @@ add_custom_target(version ALL
...
-        [ -d .git ] && git describe --always || echo ${G4CMP_VERSION}) > ${version_FILE}"
+        [ -d .git ] && git describe --always || echo ${G4CMP_USER_VERSION}) > ${version_FILE}"
     )
 
 install(FILES ${PROJECT_BINARY_DIR}/${version_FILE} DESTINATION share/G4CMP
         COMPONENT config)
 
+# Load G4CMP version into cmake flag
+if (NOT DEFINED G4CMP_VERSION)	# Only if user did not specify
+    file(READ ${PROJECT_BINARY_DIR}/${version_FILE} VERSION_CONTENT)
```
szatschler: This line may fail if this is the first time building the package from scratch as it requires to read the version text file which is only created at build time. I will need to revise the logic here.

szatschler: This has been addressed with the latest commit.

CMakeLists.txt
```diff
@@ -78,45 +79,70 @@ if (BUILD_G4CMP_TESTS)
...
-set(version_FILE .g4cmp-version)
-
-find_program(SH sh HINTS /bin)
-add_custom_target(version ALL
-    COMMENT Building version identification file
-    VERBATIM			# Allows for parentheses and redirection below
-    COMMAND ${CMAKE_COMMAND} -E env ${SH}
-        -c "(cd ${PROJECT_SOURCE_DIR};\
-        [ -d .git ] && git describe --always || echo ${G4CMP_VERSION}) > ${version_FILE}"
+set(G4CMP_VERSION ${G4CMP_USER_VERSION})
```
kelseymh: Just a question, why do it this way, rather than having the user pre-set $G4CMP_VERSION directly?  I'm sure there's a good reason, I'm just curious.

CMakeLists.txt
```diff
@@ -78,45 +79,70 @@ if (BUILD_G4CMP_TESTS)
...
         DESTINATION cmake)
-install(FILES ${PROJECT_SOURCE_DIR}/cmake/UseG4CMP.cmake
+install(FILES ${PROJECT_BINARY_DIR}/UseG4CMP.cmake
         DESTINATION cmake)
 
 #-----------------------------------------------------------------------------
 # Install example applications if requested
 #
-if(INSTALL_EXAMPLES)
+if (INSTALL_EXAMPLES)
```
kelseymh, line 132: I'm not going to ask you to change anything.  I think a long time ago we were using "if(...)" because CMake treats "if" as just another function call with a return value (c.f. the "endif()" function below).

szatschler: Right, I only wanted to match the style of the other if clauses throughout the file.

CMakeLists.txt
szatschler: Good point. I had earlier introduced the G4CMP_USER_VERSION flag to clearly distinguish the two cases when capturing the G4CMP version in a backwards compatible way (i.e. generating the .g4cmp-version file). Let me simplify this to only make use of the G4CMP_VERSION variable.

kelseymh: Ah, that makes sense.  Now you're using Git directly.

szatschler: Resolved with latest commit.

kelseymh (approved):

PR #100, opened 2026-08-28
G4CMP-663: Phonon example clean-up

BirgitZatschler:
I started an attempt to clean-up the Phonon example and in particular update the README.


examples/phonon/src/PhononDetectorConstruction.cc
```diff
@@ -16,6 +14,9 @@
 // 20251116  G4CMP-539 -- Use UpdateMPT wrapper function to set properties.
 // 20251117  G4CMP-541 -- For G4 v11, replace ::Invisible w/::GetInvisible()
 // 20260816  G4CMP-657 -- Fix typo in specCoeffs vector.
+// 20260827  B. Zatschler -- Add new function ConstructSDandField().
```
szatschler, line 17: Should we adopt the same style here (citing the G4CMP Jira issue number) like in previous header comments? It would also be nice to adopt the same style in the corresponding .hh header file.

szatschler (commented):
Very nice update! I only have one (more general) comment about the consistency of reporting in-situ code changes.

examples/phonon/src/PhononDetectorConstruction.cc
BirgitZatschler: Indeed, the comments are inconsistent. I copied it from the header file, which uses names, not the Jira ticket number. The author's name is part of the commit, so there is no real need to repeat it. The Jira ticket number seems more useful to me.

I noticed that GitHub's default tab size is 4, which displays the comment block above wrongly, because the original author set their editor to a tab size of 8. We also see this further down in the code. So if people copy this and have their own editor set to a tab size of e.g. 2, you may get completely inconsistent results, e.g. [here](https://github.com/G4CMP/G4CMP/blob/caa143549a5f52d8f0c85b7532cad2068d1f7c95/examples/heterostructure/src/HeterostructureDetectorConstruction.cc#L154-L172).

@kelseymh What is your preference?
I can replace the tabs with the correct number of spaces easily, taking into account that a tab is not always 8 spaces, but moves the cursor to the next position at 8, 16, ...

kelseymh: If you want to change it to the ticket number, rather than name, I am fine with that. Or not. You'll see that we are inconsistent about it throughout the code base :-(

Tabs at columns modulo 8 is the ASCII standard for TAB (0x09). Take a look at it in your terminal window. The mapping to different columations is specific to a variety of graphical editors, and can be set as a user preference in each of them. If you want to convert them to spaces, that would be terrific.

examples/phonon/src/PhononSensitivity.cc
```diff
@@ -91,10 +76,10 @@ void PhononSensitivity::SetOutputFile(const G4String &fn) {
...
       output.close();
     } else {
-      output << "Run ID,Event ID,Track ID,Particle Name,Start Energy [eV],"
-             << "Start X [m],Start Y [m],Start Z [m],Start Time [ns],"
-             << "Energy Deposited [eV],Track Weight,End X [m],End Y [m],End Z [m],"
-             << "Final Time [ns]\n";
+      output << "Run_ID,Event_ID,Track_ID,Particle_Name,Start_Energy_[eV],"
+             << "Start_X_[m],Start_Y_[m],Start_Z_[m],Start_Time_[ns],"
+             << "Energy_Deposited_[eV],Track_Weight,End_X_[m],End_Y_[m],End_Z_[m],"
+             << "Final_Time_[ns]\n";
```
kelseymh, line 82: Thank you for doing this!  Is your thought to do the same in all the other examples?

kelseymh (approved):

examples/phonon/src/PhononDetectorConstruction.cc
BirgitZatschler: I changed the comment to use the ticket number instead of my names. I did not go through the tedious task to do that for all existing entries in the header.

I replaced all tabs in the phonon example with spaces taking into account that tabs are at columns modulo 8. A few lines needed manual intervention, because the indention looked wrong.

examples/phonon/src/PhononConfigMessenger.cc
BirgitZatschler, line 21: Mike pointed out in another PR:

>  The G4CMP examples that use a ConfigManager all keep the "/g4cmp/" prefix (which sort of makes sense). But that creates problems if the same command (like "verbose") is defined in multiple Messengers.

Would you like to stick to this and keep the `/g4cmp/` prefix here or do you prefer something different as a hint to a user who takes the phonon example as a base of another example or application to change to to something else?

examples/phonon/src/PhononSensitivity.cc
BirgitZatschler: Eventually, we can change this in all other examples if we want. I guess that I would start a similar clean-up for each example separately.

examples/phonon/src/PhononConfigMessenger.cc
kelseymh: I'd be nervous about changing that now, just because it's been like this a long time, the included macro files use it (I think), and anyone will their own macros will also use it (c.f. G4CMP-668 :-( ).

BirgitZatschler: The macro files in the phonon example do not use the one macro command the phonon example provides. I don't see a branch for G4CMP-668, so I can't comment on that. Anyway, I'm fine not touching this to not break anything.

kelseymh (approved):

PR #102, opened 2026-09-15
G4CMP-662: Modifying QPBoundaryProcess to adaptively set QP reflection coefficient to reduce bias from WoS at transparent SC-SC interfaces

relineha:
Have modified the QPBoundaryProcess functions to counter bias introduced from applying the Walk-on-spheres algorithm at transparent superconductor-superconductor (SC-SC) boundaries. The strategy is to adaptively change the QP reflection parameter based on local geometry and local race-winning physics rates on both sides of the boundary. Link to upcoming presentation (will be finished as of 9/17/26): https://docs.google.com/presentation/d/1De5Av_GabH2ZpndeiDd-Asxcj8mtrpcsElNM-nV3cI0/edit?slide=id.p#slide=id.p


relineha:
Recognize there are quite a few outstanding issues/things needed cleaning up here. Will handle in the next few days and then request a code review.

BirgitZatschler:
It looks like branch G4CMP-662 was accidentally branched off of master. Please keep in mind to branch off of develop in the future: [G4CMP Code Management](https://confluence.slac.stanford.edu/spaces/G4CMP/pages/532889793/G4CMP+Code+Management#G4CMPCodeManagement-DevelopmentorFeatureBranches)

library/include/G4CMPBoundaryUtils.hh
```diff
@@ -23,6 +23,8 @@
 // 20251028  G4CMP-527: Move CheckStepBoundary() here from DriftBoundaryProcess
 // 20251204  G4CMP-511 -- Create parallel Lambertian reflection code for charges.
 // 20251210  G4CMP-518 -- Make PhononVelocityIsInward() generic.
+// 20260826  G4CMP-662: Make ReflectTrack() non-const so that QPs
+//       can update their lattice while determining reflection probability
```
BirgitZatschler: Can you properly indent this line? The same applies to several other change comments in file headers.

library/include/G4CMPDriftBoundaryProcess.hh
```diff
@@ -77,6 +77,7 @@ protected:
   virtual void DoFinalReflection(const G4Track& aTrack,const G4Step& aStep,
 				 G4ParticleChange& aParticleChange);
 
+
```
BirgitZatschler, line 80: I'm certain Mike is going to request to remove this extraneous line.
Before adding files to a commit you can check for changes with `git diff <filename>` to catch such changes and if applicable restore the files and don't commit them. Same applies to library/include/G4CMPPhononBoundaryProcess.hh where just 2 spaces were added.

library/src/G4CMPBoundaryUtils.cc
```diff
@@ -796,41 +796,56 @@ void
 G4CMPBoundaryUtils::ApplyBoundaryAction(const G4Track& aTrack,
 					const G4Step& aStep,
 					G4ParticleChange& aParticleChange) {
+  //G4cout << "ABA: Function Point 0" << G4endl;
```
BirgitZatschler: Can you remove this and similar comments - or if applicable convert them to proper verbosity output?

library/src/G4CMPBoundaryUtils.cc
```diff
@@ -859,18 +875,34 @@ G4bool G4CMPBoundaryUtils::AbsorbTrack(const G4Track&, const G4Step&) const {
...
+//quasi-transparent) boundaries.
+G4bool G4CMPBoundaryUtils::ReflectTrack(const G4Track& aTrack,
+                                        const G4Step&) {
+  G4Exception("G4CMPBoundaryUtils::ReflectTrack","Boundary015",FatalException,
+              "Non-const base class ReflectTrack() is running? Problematic!");
+  return false;
+}
+*/
+
 G4bool G4CMPBoundaryUtils::MaximumReflections(const G4Track& aTrack) const {
```
BirgitZatschler, line 874: Do you want to remove this or actually overload the function?

library/src/G4CMPQPBoundaryProcess.cc
```diff
@@ -245,22 +256,362 @@ G4CMPQPBoundaryProcess::PostStepDoIt(const G4Track& aTrack,
...
+    preVolRates[3] = dynamic_cast<G4CMPQPLocalTrappingProcess*>
+      (pl[relevantQPProcessIDs[3]])->GetRateModel()->Rate(aTrack);
+  }
+
+  
+  //Do the transition to the new lattice (used to be in DoTransmission)
+  //Since the lattice hasn't changed yet, change it here. (This also happens
+  //at the MFP calc point at the beginning of the next step, but it's nice to
+  //have it here so we can use the new lattice info to help figure out vdir,
+  //etc.) REL SHOULD CHECK TO MAKE SURE THIS DOES NOT CAUSE WEIRD BEHAVIOR
```
BirgitZatschler: I just leave this comment here as a reminder to check what is pointed out in this comment.

library/src/G4CMPQPBoundaryProcess.cc
```diff
@@ -245,22 +256,362 @@ G4CMPQPBoundaryProcess::PostStepDoIt(const G4Track& aTrack,
...
+         << "\n, postStepSafety: " << postStepSafety
+         << "\n, pre-vol qpRadiatesPhonon rate: " << preVolRates[0]
+         << "\n, post-vol qpRadiatesPhonon rate: " << postVolRates[0]
+         << "\n, pre-vol qpRecombination rate: " << preVolRates[1]
+         << "\n, post-vol qpRecombination rate: " << postVolRates[1]
+         << "\n, pre-vol qpDiffusionTimeStepper rate: " << preVolRates[2]
+         << "\n, post-vol qpDiffusionTimeStepper rate: " << postVolRates[2]
+         << "\n, pre-vol qpLocalTrapping rate: " << preVolRates[3]
+         << "\n, post-vol qpLocalTrapping rate: " << postVolRates[3]    
+         << G4endl;
```
BirgitZatschler: Would you like to include this verbosity output?

library/src/G4CMPQPBoundaryProcess.cc
```diff
@@ -362,14 +716,20 @@ void G4CMPQPBoundaryProcess::DoTransmission(const G4Track& aTrack,
...
+  /*  
+  // THIS IS NOW DONE IN REFLECTTRACK
   //Since the lattice hasn't changed yet, change it here. (This also happens
   //at the MFP calc point at the beginning of the next step, but it's nice to
   //have it here so we can use the new lattice info to help figure out vdir,
   //etc.)
   this->SetLattice(G4LatticeManager::GetLatticeManager()->GetLattice(aStep.GetPostStepPoint()->GetPhysicalVolume()));
   UpdateSCAfterLatticeChange();
-    
+  */
```
BirgitZatschler: Remove the code if you don't need it anymore.

library/src/G4CMPQPRadiatesPhononRate.cc
```diff
@@ -6,33 +6,86 @@
...
+      G4Exception("G4CMPQPRadiatesPhononRate::Rate",
+                  "QPRadiatesPhononRate000",FatalException, msg);
+      return 0;
+    }
+    //G4cout << "The count, " <<
+    //  "fMap_physicalLattice_NormalizedTauQPRadiatesPhononVsEnergy.count(theLat): "
+    //       <<
+    //  fMap_physicalLattice_NormalizedTauQPRadiatesPhononVsEnergy.count(theLat)
+    //       << G4endl;
+    tau_scattering = theLat->GetSCTau0qp()*
```
BirgitZatschler, line 77: Convert to verbosity output?

library/src/G4CMPQPRecombinationRate.cc
```diff
@@ -16,29 +20,76 @@
...
+}
 
 // Recombination rate is computed using energy and the G4SCUtils class, upon
-// which this is based
-G4double G4CMPQPRecombinationRate::Rate(const G4Track& aTrack) const
-{
+// which this is based. If there is an additional lattice provided, then compute
+// the rate based on that lattice
+G4double G4CMPQPRecombinationRate::
+Rate(const G4Track& aTrack,const G4LatticePhysical * theLat) const {
```
BirgitZatschler: Insert a space after the comma. For consistency, I prefer having the pointer `*` attached to the data type, i.e. `G4LatticePhysical* theLat`.

library/src/G4CMPQPRecombinationRate.cc
```diff
@@ -16,29 +20,76 @@
...
+          << " versus energy. This is problematic. Fix.";
+      G4Exception("G4CMPQPRecombinationRate::Rate",
+                  "QPRecombinationRate000",FatalException, msg);
+      return 0;
+    }
+    //G4cout << "The count, " <<
+    //  "fMap_physicalLattice_NormalizedTauRecombinationVsEnergy.count(theLat): "
+    //       <<
+    //  fMap_physicalLattice_NormalizedTauRecombinationVsEnergy.count(theLat)
+    //       << G4endl;
```
BirgitZatschler: Convert this to a verbosity output?

BirgitZatschler (commented):
I pointed out a few style specific things.

library/include/G4CMPBoundaryUtils.hh
relineha: Ope, good catch. I have now indented line 26 properly, and have line 27 matching the other overflow line in the change history in this file (the one from 2017).

library/include/G4CMPDriftBoundaryProcess.hh
relineha: Yep -- I generally do this but some slip through the cracks... Done.

library/src/G4CMPBoundaryUtils.cc
relineha: Yep -- sorry. I saw these yesterday and planned to remove them today before requesting a code review, so you're a little fast on the draw :P 

Done.

library/src/G4CMPBoundaryUtils.cc
relineha: Dangling -- falls under the purview of things I caught late and was planning to remove today. Removed.

library/src/G4CMPQPBoundaryProcess.cc
relineha: Thanks. I think this got covered by recent validations I've done. Removed

library/src/G4CMPQPBoundaryProcess.cc
relineha: Returned to normal state. Apologies.

library/src/G4CMPQPBoundaryProcess.cc
relineha: Thanks for the catch. Done. Apologies again!

library/src/G4CMPQPRadiatesPhononRate.cc
relineha: Not worth it -- deleted.

library/src/G4CMPQPRecombinationRate.cc
relineha: Done and done.

library/src/G4CMPQPRecombinationRate.cc
relineha: Nah, not worth it. Removed.

relineha:
> It looks like branch G4CMP-662 was accidentally branched off of master. Please keep in mind to branch off of develop in the future: [G4CMP Code Management](https://confluence.slac.stanford.edu/spaces/G4CMP/pages/532889793/G4CMP+Code+Management#G4CMPCodeManagement-DevelopmentorFeatureBranches)

@BirgitZatschler Ack, you're right -- this was an oversight on my end that I won't repeat... Given that I did merge develop into this before submitting the pull request, is there any issue with having done this aside from the fact that it's bad practice and can get me into trouble if I, for example, forget to merge develop back in before making a pull request? Trying to figure out what the risks at the current moment are if I've successfully compiled and run this after merging develop onto it.

library/include/G4CMPBoundaryUtils.hh
relineha: Actually, since there does not seem to be a standard for where the second line indenting should go throughout the code and it is not explicitly designated in the style guide on confluence, I'm going to align to the G in G4CMP-XXX.

relineha: Actually, since this is really non-standard throughout the code, I'm going to match the second-line indentation to whatever else the format is in the file I'm in...

kelseymh:
About the "branch from master" issue.  It's far too late for you to change it now; we would lose all of the history of your development work and commits -- unless we gave you specific guidance on moving thing over one by one, and even then we'd lose the date history.  

The underlying problem with branching master -> feature -> develop is that we now no longer have separate master and develop branches.  All of the past "release merges" (develop -> master), with their commits and V-tags, are now inserted into the develop branch, where they don't belong.

As far as the source code is concerned, they're equivalent, but in terms of reconstructing history, rebasing, etc., it's a problem for Git.

library/include/G4CMPBoundaryUtils.hh
```diff
@@ -23,6 +23,8 @@
 // 20251028  G4CMP-527: Move CheckStepBoundary() here from DriftBoundaryProcess
 // 20251204  G4CMP-511 -- Create parallel Lambertian reflection code for charges.
 // 20251210  G4CMP-518 -- Make PhononVelocityIsInward() generic.
+// 20260826  G4CMP-662 -- Make ReflectTrack() non-const so that QPs
+//           can update their lattice while determining reflection probability
```
kelseymh, line 27: I'd like to understand this better.  It's not really a great system to update the lattice cached in G4CMPProcessUtils during processing.  It means you no longer have a consistent state if that function falls through to another.  Can you not get the "opposite side lattice" into a local variable and use it?

library/include/G4CMPQPRadiatesPhononRate.hh
```diff
@@ -25,9 +28,11 @@ public:
   virtual ~G4CMPQPRadiatesPhononRate() {;}
   
   virtual G4double Rate(const G4Track& aTrack) const;
+  virtual G4double Rate(const G4Track& aTrack,
+                        const G4LatticePhysical * theLat) const;
```
kelseymh, line 32: Please remove extra space before "*".  Our style guide treats "Type*" as the data type, rather than associating the "*" with the variable.

This applies throughout your code changes below.  If you'd like me to flag all of them, let me know.

library/src/G4CMPQPRadiatesPhononRate.cc
```diff
@@ -6,33 +6,81 @@
...
+
   //Debugging
   if (verboseLevel > 5) {
     G4cout << "-- G4CMPQPRadiatesPhononRate::Rate --" << G4endl;
   } 
+
+  //First, check to see if the input lattice is zero. If it is, then we're
+  //going to use the "current" lattice to do the rate calculation.
+  bool useInputLat = false;
+  if (theLat != 0) useInputLat = true;
```
kelseymh, line 38: You could do this more easily.  If "theLat" input argument is zero, you could _set it_ (the pointer is passed by value, so this doesn't change the calling code) to the "current" lattice from CDMSProcTools, and then just use "theLat" everywhere below.

Much simpler that carying a flag around.

library/src/G4CMPQPRadiatesPhononRate.cc
```diff
@@ -58,32 +106,69 @@ G4double G4CMPQPRadiatesPhononRate::Rate(const G4Track& aTrack) const
...
+CheckToSeeSCParametersSet(const G4LatticePhysical * theLat) const {
+
+  //If a lattice is provided, we want to ask about that lattice info. If one is
+  //not, we want to ask about the lattice info currently loaded in
+  if (theLat == 0) {
+
+    //Check for the gap0energy, Tcrit, Teff, and Tau0qp. If all of these aren't
+    //set, return false. However, if any subset of them are set, then throw a
+    //flag--means that someone may just have forgot one of them.
+    if (fGap0Energy==0 || fTau0_qp == DBL_MAX || fTcrit == 0 || fTeff == 0) {
```
kelseymh, line 119: Could we have a consistent style at least within individual if blocks.  Either use spaces around "==" or remove spaces around "==".

library/src/G4CMPQPRadiatesPhononRate.cc
```diff
@@ -58,32 +106,69 @@ G4double G4CMPQPRadiatesPhononRate::Rate(const G4Track& aTrack) const
...
+      } else {
+        //^Means that the material is partially set -- this is probably a
+        //mistake
+        G4ExceptionDescription msg;
+        msg << "Noticed that in the rate calculation step for the QP "
+            << "phononradiation process you've incorrectly defined or omitted "
+            << "the Gap0Energy parameter, the Tcrit parameter, the Teff "
+            << "parameter, or the Tau0qp parameter. In other words, you don't "
+            << "have enough input information in your config.txt file to run "
+            << "the phononRadiation physics correctly.";
```
kelseymh, line 134: Could these exception messages be made more compact?

library/src/G4CMPQPRecombinationRate.cc
```diff
@@ -16,29 +20,71 @@
...
+
   //Debugging
   if (verboseLevel > 5) {
     G4cout << "-- G4CMPQPRecombinationRate::Rate --" << G4endl;
   }
   
+  //First, check to see if the input lattice is zero. If it is, then we're
+  //going to use the "current" lattice to do the rate calculation.
+  bool useInputLat = false;
+  if (theLat != 0) useInputLat = true;
```
kelseymh, line 41: See my comment above.  Since "theLat" is a local variable and argument, you could just set it to point to the "current" lattice and get rid of the flag and all the if blocks.

library/src/G4CMPQPRecombinationRate.cc
```diff
@@ -125,6 +204,15 @@ G4CMPQPRecombinationRate::UpdateLookupTable(const G4LatticePhysical * theLat) {
   }  
 }
 
+bool G4CMPQPRecombinationRate::
+CheckLookupTableForLat(const G4LatticePhysical * theLat) const {
+  //1. If the lattice doesn't exist in the lattice container associated with
+  //   this process yet, throw a flag
+  if(fMap_physicalLattice_NormalizedTauRecombinationVsEnergy.count(theLat)
+     == 0){ return false; }
+  return true;
```
kelseymh, line 213: This is pointless.  You're returning the inverse of the if condition, so just write
```
return fMap_physicalLattice_NormalizedTauRecombinationVsEnergy.count(theLat)!=0;
```

kelseymh (changes requested):

library/include/G4CMPQPRadiatesPhononRate.hh
relineha: Ack, Birgit commented on this and this slipped through the filter. Will fix this.

library/src/G4CMPQPRadiatesPhononRate.cc
relineha: Yep, will change.

PR #103, opened 2026-09-21
G4CMP-665 Fix memory leak in G4CMPParticleChangeForPhonon

BirgitZatschler:
These code changes address the memory leak caused by the of `G4TouchableHandle theTouchableHandle` in G4CMPParticleChangeForPhonon. This data member has been removed from the class and corresponding functions have been updated accordingly.

### DIFF
--- a/library/include/G4CMPParticleChangeForQPDiffusion.hh
+++ b/library/include/G4CMPParticleChangeForQPDiffusion.hh
@@ -7,6 +7,7 @@
 /// \brief Definition of the G4CMPParticleChangeForQPDiffusion class
 //
 //  20260107  G4CMP-563: Delete copy operations for G4 v11 compatibility.
+//  20260901  G4CMP-655: Remove unnecessary include
 
 #ifndef G4CMPParticleChangeForQPDiffusion_h
 #define G4CMPParticleChangeForQPDiffusion_h 1
@@ -15,7 +16,6 @@
 #include "G4ios.hh"
 #include "G4ThreeVector.hh"
 #include "G4ThreeVector.hh"
-#include "G4TouchableHandle.hh"
 #include "G4VParticleChange.hh"
 
 class G4DynamicParticle;
--- a/library/src/G4CMPParticleChangeForPhonon.cc
+++ b/library/src/G4CMPParticleChangeForPhonon.cc
@@ -13,6 +13,7 @@
 // 20251116 For G4 11, explicitly remove the copy operators to match base.
 // 20251128 Discard touchable contents after updating.
 // 20260212 G4CMP-585 Only discard touchable if it was modified (fix mem leak)
+// 20260901 G4CMP-655: Fix memory leak caused by theTouchableHandle.
 
 #include "G4CMPParticleChangeForPhonon.hh"
 #include "G4VTouchable.hh"
@@ -27,8 +28,6 @@
 
 void G4CMPParticleChangeForPhonon::Initialize(const G4Track& track) {
   updateVol = false;
-  theTouchableHandle = track.GetTouchableHandle();
-  
   G4ParticleChange::Initialize(track);
 }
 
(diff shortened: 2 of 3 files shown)


library/src/G4CMPProcessUtils.cc
```diff
@@ -64,6 +64,7 @@
 // 20251007  Bug fix for G4CMP-497 fix just above.
 // 20260110  G4CMP-567:  Add missing else-IsHole() block in GetKineticEnergy.
 // 20260120  G4CMP-558: Add array dimension in Get*VelocityVector().
+// 20260921  G4CMP-655: Change ProposeNewTouchableHandle to take no argument.
```
szatschler, line 67: Should this be G4CMP-665 instead of 655?

library/include/G4CMPParticleChangeForPhonon.hh
```diff
@@ -11,9 +11,11 @@
 //
 // 20250410 Implement ParticleChange for phonons to handle displaced reflections
 // 20250413 Add Initialize() implementation to reset updateVol flag, add
-//		missing copy operations, may be needed
+//          missing copy operations, may be needed
 // 20251116 For G4 11, explicitly remove the copy operators to match base.
 // 20251128 Implement empty destructor to avoid deleting G4TouchableHandle.
+// 20260921 G4CMP-655: Remove G4TouchableHandle to avoid memory leak and seg
```
szatschler, line 17: I think this should be G4CMP-665.

library/include/G4CMPParticleChangeForPhonon.hh
```diff
@@ -38,20 +40,14 @@ public:
   G4Step* UpdateStepForPostStep(G4Step* pStep) final;
   
   // --- Methods for proposing PostStep volume ---
-  void ProposeTouchableHandle(const G4TouchableHandle& nextTouchableHandle) {
-    theTouchableHandle = nextTouchableHandle;
+  void ProposeNewTouchableHandle() {
```
szatschler, line 43: Is this still the best function name? It may be a bit misleading moving forward.

library/include/G4CMPParticleChangeForQPDiffusion.hh
```diff
@@ -7,6 +7,7 @@
 /// \brief Definition of the G4CMPParticleChangeForQPDiffusion class
 //
 //  20260107  G4CMP-563: Delete copy operations for G4 v11 compatibility.
+//  20260901  G4CMP-655: Remove unnecessary include
```
szatschler, line 10: Should this be G4CMP-665?

library/src/G4CMPParticleChangeForPhonon.cc
```diff
@@ -13,6 +13,9 @@
 // 20251116 For G4 11, explicitly remove the copy operators to match base.
 // 20251128 Discard touchable contents after updating.
 // 20260212 G4CMP-585 Only discard touchable if it was modified (fix mem leak)
+// 20260901 G4CMP-655: Fix memory leak caused by theTouchableHandle.
```
szatschler, line 16: Should this line be removed in favor of just one line?

library/src/G4CMPParticleChangeForPhonon.cc
```diff
@@ -13,6 +13,9 @@
...
+// 20260921 G4CMP-655: Remove G4TouchableHandle to avoid memory leak and seg
```
szatschler, line 17: Should this be G4CMP-665?

library/src/G4CMPProcessUtils.cc
```diff
@@ -259,7 +260,7 @@ void G4CMPProcessUtils::FillParticleChange(G4CMPParticleChangeForPhonon& particl
 
     // Update position, touchable, and step status
     particleChange.ProposePosition(position);
-    particleChange.ProposeTouchableHandle(step.GetPreStepPoint()->GetTouchableHandle());
+    particleChange.ProposeNewTouchableHandle();
```
szatschler, line 263: See previous comment. Maybe we can find a better fitting function name?

szatschler (commented):
I have a few stylistic comments and suggestions. See the details in the review.

library/include/G4CMPParticleChangeForPhonon.hh
BirgitZatschler: Thanks for catching this. Looks like I messed it up once and copied & pasted it to be consistently wrong in every file. Fixed.

library/include/G4CMPParticleChangeForQPDiffusion.hh
BirgitZatschler: Fixed.

library/src/G4CMPParticleChangeForPhonon.cc
BirgitZatschler: I'm usually trying to update the header for every commit, so it is clear to which issue the commit belongs to. Has @kelseymh any preferences?

library/src/G4CMPParticleChangeForPhonon.cc
BirgitZatschler: Fixed.

library/src/G4CMPProcessUtils.cc
BirgitZatschler: Fixed.

library/include/G4CMPParticleChangeForPhonon.hh
BirgitZatschler: Fair point. The function only sets the boolean `updateVol` to `true`. If it is `true`, the code in `G4CMPParticleChangeForPhonon::UpdateStepForPostStep()` sets the Touchable Handle of the post step point to be the same as in the pre step point.
So maybe a better function name could be `UpdateTouchableHandle`?

